PR #1, opened 2022-01-06
DI Setup

cristhianescobar:
To create a solid and extensible application that scales to the needs of projectNEWM we can use Hilt as the DI tool to manage dependencies.


andymiller813 (approved):

newmskywalker (approved):
💯

PR #3, opened 2022-01-13
UI Android Compose

cristhianescobar:
As per previous conversations, we will be migrating the UI that we currently have to compose.

On this PR, I'm adding dependencies needed plus an entry point to a compose activity where we will be getting away from normal android UI xmls and adopting composables through the app.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -28,10 +28,21 @@ android {
     }
 
     kotlinOptions {
-        jvmTarget = "1.8"
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
+        useIR = true
     }
     buildFeatures {
+        compose = true
         viewBinding = true
+    }
+    composeOptions {
+        kotlinCompilerExtensionVersion = "1.0.4"
+    }
+
+    packagingOptions {
+        resources {
+            excludes += "/META-INF/{AL2.0,LGPL2.1}"
+        }
     }
 }
 
@@ -44,13 +55,21 @@ dependencies {
     implementation(Deps.androidxNavigationFragmentKtx)
     implementation(Deps.androidxNavigationUiKtx)
     implementation(Deps.androidxSplashScreen)
-    implementation(Deps.kotlinStdLib)
     implementation(Deps.material)
     implementation(Deps.viewBindingPropertyDelegate)
     implementation(Deps.androidHilt)
     kapt(Deps.androidHiltCompiler)
     kapt(Deps.hiltCompiler)
+    implementation("com.google.android.material:material:1.4.0")
+    implementation("androidx.compose.ui:ui:1.0.5")
+    implementation("androidx.compose.material:material:1.0.5")
+    implementation("androidx.compose.ui:ui-tooling-preview:1.0.5")
+    implementation("androidx.lifecycle:lifecycle-runtime-ktx:2.4.0")
+    implementation("androidx.activity:activity-compose:1.4.0")
+    implementation("androidx.navigation:navigation-compose:2.4.0-rc01")
 
+    androidTestImplementation("androidx.compose.ui:ui-test-junit4:1.0.5")
+    debugImplementation("androidx.compose.ui:ui-tooling:1.0.5")
     testImplementation(Deps.junit)
 
     androidTestImplementation(Deps.androidxJUnit)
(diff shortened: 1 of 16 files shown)


androidApp/build.gradle.kts
```diff
@@ -44,13 +55,21 @@ dependencies {
...
+    implementation("com.google.android.material:material:1.4.0")
+    implementation("androidx.compose.ui:ui:1.0.5")
+    implementation("androidx.compose.material:material:1.0.5")
+    implementation("androidx.compose.ui:ui-tooling-preview:1.0.5")
+    implementation("androidx.lifecycle:lifecycle-runtime-ktx:2.4.0")
+    implementation("androidx.activity:activity-compose:1.4.0")
+    implementation("androidx.navigation:navigation-compose:2.4.0-rc01")
 
+    androidTestImplementation("androidx.compose.ui:ui-test-junit4:1.0.5")
+    debugImplementation("androidx.compose.ui:ui-tooling:1.0.5")
```
cristhianescobar, line 72: Will clean up dependencies once we removed the non compose dependencies.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -3,13 +3,13 @@ import org.gradle.api.artifacts.dsl.RepositoryHandler
 @Suppress("unused")
 object Versions {
     const val compileSdkVersion = 31
-    const val minSdkVersion = 23
+    const val minSdkVersion = 25
```
newmskywalker, line 6: Wondering if this min sdk version is too high. Maybe we can look at what percentage of devices we would lose with this bump

newmskywalker (approved):
Lgtm!

androidApp/src/main/java/com/projectnewm/ui/theme/Color.kt
```diff
@@ -0,0 +1,8 @@
+package com.projectnewm.ui.theme
+
+import androidx.compose.ui.graphics.Color
+
+val Purple200 = Color(0xFFBB86FC)
```
andymiller813, line 5: Oh man even colors and dimensions in Kotlin.  What a time to be alive

androidApp/src/main/java/com/projectnewm/ui/theme/Type.kt
```diff
@@ -0,0 +1,28 @@
...
+import androidx.compose.material.Typography
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontFamily
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.sp
+
+// Set of Material typography styles to start with
+val Typography = Typography(
+    body1 = TextStyle(
+        fontFamily = FontFamily.Default,
```
andymiller813, line 12: Where is it getting the default from? XML theme?

buildSrc/src/main/java/Dependencies.kt
andymiller813: It's probably fine, marshmallow is 3.5%, Lollipop 2.9% and Kitkat 1%.  By the time we actually release this those will be even lower and/or we can revisit later

andymiller813 (approved):

androidApp/src/main/res/values/themes.xml
```diff
@@ -36,4 +36,13 @@
         <item name="windowSplashScreenAnimatedIcon">@mipmap/ic_launcher</item>
         <item name="postSplashScreenTheme">@style/Theme.ProjectNewm.NoActionBar</item>
     </style>
+
+    <style name="Theme.ProjectNewm.SplashScreen.NoActionBar">
+        <item name="windowActionBar">false</item>
```
andymiller813, line 41: Do we want to just add these to `Theme.ProjectNewm.SplashScreen`? I wouldn't think we will ever need both

andymiller813: BTW this is what switches the theme after splash: 
```<item name="postSplashScreenTheme">@style/Theme.ProjectNewm.NoActionBar</item>```

buildSrc/src/main/java/Dependencies.kt
cristhianescobar: I thought about it but don't think it will make a big difference by the time it actually gets release on the play store

androidApp/src/main/java/com/projectnewm/ui/theme/Type.kt
cristhianescobar: Compose seems to have default values based on the Material theme

androidApp/src/main/res/values/themes.xml
cristhianescobar: I can make that part of the next set of changes 👍🏼

PR #5, opened 2022-01-17
Setting up Bottom Navigation

cristhianescobar:
Setting up bottom bar navigation based on current design


androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
```diff
@@ -0,0 +1,90 @@
...
+            onClick = { onNavigationSelected(Screen.HomeRoot) }
+        )
+        HomeBottomNavigationItem(
+            icon = R.drawable.ic_dashboard_black_24dp,
+            selected = currentRootScreen == Screen.TribeRoot,
+            label = stringResource(R.string.tribe),
+            onClick = { onNavigationSelected(Screen.TribeRoot) }
+        )
+        HomeBottomNavigationItem(
+            icon = R.drawable.ic_notifications_black_24dp,
```
andymiller813: we should be able to get the proper icons now but we can do that as a follow up

androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
```diff
@@ -0,0 +1,90 @@
...
+import androidx.navigation.NavHostController
+import androidx.navigation.compose.rememberNavController
+import com.projectnewm.R
+import kotlinx.coroutines.flow.collect
+
+@Composable
+internal fun NEWMApp(navController: NavHostController = rememberNavController()) {
+    Scaffold(
+        bottomBar = {
+            val currentRootScreen by navController.currentRootScreenAsState()
```
andymiller813, line 19: Is there any way to hoist this state up a level so it gets remembered and we don't have to go find the entry tab each time?

androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
```diff
@@ -0,0 +1,90 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.navigation.NavController
+import androidx.navigation.NavHostController
+import androidx.navigation.compose.rememberNavController
+import com.projectnewm.R
+import kotlinx.coroutines.flow.collect
+
+@Composable
+internal fun NEWMApp(navController: NavHostController = rememberNavController()) {
```
andymiller813: nit: 
```suggestion
internal fun NewmApp(navController: NavHostController = rememberNavController()) {
```

androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
```diff
@@ -0,0 +1,90 @@
...
+            NEWMBottomNavigation(
+                currentRootScreen = currentRootScreen,
+                onNavigationSelected = { navController.navigate(it.route) }
+            )
+        }
+    ) { }
+}
+
+@Composable
+internal fun NEWMBottomNavigation(
```
andymiller813: nit: 
```suggestion
internal fun NewmBottomNavigation(
```

andymiller813 (approved):

androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
cristhianescobar: I added the icons on the last commit, however they seem to need a bit of work from design, they don't scale well on the app on their current form

androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
cristhianescobar: We can, the next PR that sets up the nav flow will have it, good call out

androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
cristhianescobar: Change made 👍🏼

androidApp/src/main/java/com/projectnewm/compose/NEWMApp.kt
cristhianescobar: Change made 👍🏼

PR #7, opened 2022-01-17
Connect Bottom Nav with Screens

cristhianescobar:
Adding the ability to navigate through the different screens, Home, Tribe, Stars, Wallet

and also adding an example how to navigate to details page


androidApp/src/main/java/com/projectnewm/compose/screens/HomeScreen.kt
```diff
@@ -29,5 +29,9 @@ fun HomeScreen() {
             textAlign = TextAlign.Center,
             fontSize = 25.sp
         )
+
+        Button(onClick = {onShowDetails(3)}, ) {
```
andymiller813, line 33: nit: Formatting looks off here

andymiller813 (approved):

PR #9, opened 2022-01-19
Build debug apk on development push

newmskywalker:
Deploying the debug apk to firebase distribution on push to the development branch.

Considered doing it for every PR, but not sure how useful that would be

### DIFF
--- a/.github/workflows/android-preview-pr.yml
+++ b/.github/workflows/android-preview-pr.yml
@@ -1,0 +1,35 @@
+on:
+  push:
+    branches:
+      - development
+    paths:
+     - "androidApp/**"
+     - "shared/**"
+     - "buildSrc/**"
+     - ".github/workflows/android-preview-pr.yml"
+jobs:
+  build_android_debug:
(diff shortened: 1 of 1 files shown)


andymiller813 (approved):
Awesome!

andymiller813:
> Deploying the debug apk to firebase distribution on push to the development branch.
> 
> Considered doing it for every PR, but not sure how useful that would be

If we do it on every PR we would make sure that this step doesn't fail after merge

cristhianescobar (approved):
Very niceeeee! 💯  👍🏼

newmskywalker:
> > Deploying the debug apk to firebase distribution on push to the development branch.
> 
> > 
> 
> > Considered doing it for every PR, but not sure how useful that would be
> 
> 
> 
> If we do it on every PR we would make sure that this step doesn't fail after merge 

That's a good call out. I'm adding another check to build and run the test suite on every PR, so hopefully we'll be able to catch broken builds there. If we do still run into issues with broken builds after merge, I can turn this on for every PR

PR #12, opened 2022-01-23
Add tests for iOS bottom nav bar taps

newmskywalker:
This is my first time writing iOS tests, so there are probably better ways to do this 😅

One thing I haven't been able to figure out yet is how to test that the view is being shown, not 
necessarily what's in it. Is there a good way to do that? Maybe finding it by struct type?

I just wanted to add some basic tests so I can start writing some workflows for testing, etc.﻿


iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
newmskywalker, line 1: is this even needed?

iosApp/iosApp/StarsView.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  TribeView.swift
```
martyu:
```suggestion
//  StarsView.swift
```

iosApp/iosApp/StarsView.swift
```diff
@@ -0,0 +1,21 @@
...
+//  iosApp
+//
+//  Created by Jose Mateo on 1/23/22.
+//  Copyright © 2022 orgName. All rights reserved.
```
martyu: @andymiller813 doesn't think we need the copyright line, I've been excluding them 🤷

iosApp/iosApp/WalletView.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  TribeView.swift
```
martyu:
```suggestion
//  WalletView.swift
```

iosApp/iosApp/WalletView.swift
```diff
@@ -0,0 +1,21 @@
...
+//  iosApp
+//
+//  Created by Jose Mateo on 1/23/22.
+//  Copyright © 2022 orgName. All rights reserved.
```
martyu: Same here

iosApp/iosAppUITests/bottomNavigationUITests.swift
```diff
@@ -0,0 +1,46 @@
...
+//  iosAppUITests
+//
+//  Created by Jose Mateo on 1/23/22.
+//  Copyright © 2022 orgName. All rights reserved.
+//
+
+import XCTest
+import iosApp
+
+class bottomNavigationUITests: XCTestCase {
```
martyu: Swift naming convention for classes is Capitalized. (Same for file name)

```suggestion
class BottomNavigationUITests: XCTestCase {
```

iosApp/iosAppUITests/bottomNavigationUITests.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  iosAppUITests.swift
+//  iosAppUITests
+//
+//  Created by Jose Mateo on 1/23/22.
+//  Copyright © 2022 orgName. All rights reserved.
```
martyu: Same here

iosApp/iosAppUITests/bottomNavigationUITests.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  iosAppUITests.swift
```
martyu:
```suggestion
//  BottomNavigationUITests.swift
```

martyu:
```suggestion
//  BottomNavigationUITests.swift
```

iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
martyu: I assume this file was auto-added by xcode

martyu (changes requested):
I'm not too familiar with UI automation testing on iOS.  I think we'll want to focus on unit tests at first, since the UI is currently in flux.  But since navigation is based on the state of the View Model, you could use that to test tab bar navigation.  TBH I'm not sure what the best way to test this would be.  My guess is you could set MainViewModel.selectedTab = .home, and then call MainView.body and see if the object returned is the view type you'd want?  But also, there's an issue with the SwiftUI tab bar where the "More" tab that gets auto-added when there's more than 5 tabs doesn't work right now.  I need to figure out how we're going to handle that, which may break some of these UI tests.

iosApp/iosApp/StarsView.swift
newmskywalker: I'll remove it and tune my XC settings

iosApp/iosAppUITests/bottomNavigationUITests.swift
newmskywalker: Agree I'll update these. It was weird that Xcode autogenerated the test  files with a lower case letter before I renamed it

newmskywalker:
> I'm not too familiar with UI automation testing on iOS. I think we'll want to focus on unit tests at first, since the UI is currently in flux. But since navigation is based on the state of the View Model, you could use that to test tab bar navigation. TBH I'm not sure what the best way to test this would be. My guess is you could set MainViewModel.selectedTab = .home, and then call MainView.body and see if the object returned is the view type you'd want? But also, there's an issue with the SwiftUI tab bar where the "More" tab that gets auto-added when there's more than 5 tabs doesn't work right now. I need to figure out how we're going to handle that, which may break some of these UI tests.

I agree that most of our tests should be unit tests, but UI tests play an important role because there's usually a gap between the view model and the view when it comes to testing. Something as simple as the view not getting the updated state, or views missing completely could go wrong and it would be hard to find it early without UI tests. 

I decided to start with UI tests because those are generally more challenging to run on CI, so I wanted to get the "hard" part out of the way 😄. Once we start adding some unit tests, I'll go ahead and add some automations around them. I'll keep an eye out for those!

 As things change, hopefully the tests will evolve with them and give us more confidence to deploy changes

newmskywalker:
@martyu I made the changes you suggested. Let me know what you think!

iosApp/iosApp/Login/LoginView.swift
```diff
@@ -70,6 +70,7 @@ struct LoginView: View {
 			.background(LinearGradient(colors: [.orange, .red], startPoint: .top, endPoint: .bottom))
 			.cornerRadius(10)
 			.padding(.bottom)
+            .accessibilityIdentifier("enterNewmButton")
```
newmskywalker, line 73: not sure why it's showing up like this here, but looks fine in XC

AE542:
> This is my first time writing iOS tests, so there are probably better ways to do this 😅
> 
> One thing I haven't been able to figure out yet is how to test that the view is being shown, not necessarily what's in it. Is there a good way to do that? Maybe finding it by struct type?
> 
> I just wanted to add some basic tests so I can start writing some workflows for testing, etc.

Hey I read your tests and they do as intended! I'm really new to OpenSource but I was wondering if I could also make a commit for the tests as I'd like to try and write some.

Testing that the view is being shown is fairly simple with UIKit but I'm not sure about SwiftUI. Did you make any headway on that?

iosApp/iosApp/Login/LoginView.swift
martyu: Looks fine to me... 🤷 
image

martyu (approved):

j-mateo:
> Details

Hey! Thanks for taking the time to look it over. For now I left it as is, but I'm looking into doing screenshot tests instead. 

As for writing tests, I don't think we've defined the process for external contributions yet. Let me check with the team to see what the best way to move forward would be

newmskywalker:
> > This is my first time writing iOS tests, so there are probably better ways to do this 😅
> > One thing I haven't been able to figure out yet is how to test that the view is being shown, not necessarily what's in it. Is there a good way to do that? Maybe finding it by struct type?
> > I just wanted to add some basic tests so I can start writing some workflows for testing, etc.
> 
> Hey I read your tests and they do as intended! I'm really new to OpenSource but I was wondering if I could also make a commit for the tests as I'd like to try and write some.
> 
> Testing that the view is being shown is fairly simple with UIKit but I'm not sure about SwiftUI. Did you make any headway on that?

Hey! Thanks for taking the time to look it over. For now I left it as is, but I'm looking into doing screenshot tests instead.

As for writing tests, I don't think we've defined the process for external contributions yet. Let me check with the team to see what the best way to move forward would be

AE542:
> > > This is my first time writing iOS tests, so there are probably better ways to do this 😅
> > > One thing I haven't been able to figure out yet is how to test that the view is being shown, not necessarily what's in it. Is there a good way to do that? Maybe finding it by struct type?
> > > I just wanted to add some basic tests so I can start writing some workflows for testing, etc.
> > 
> > 
> > Hey I read your tests and they do as intended! I'm really new to OpenSource but I was wondering if I could also make a commit for the tests as I'd like to try and write some.
> > Testing that the view is being shown is fairly simple with UIKit but I'm not sure about SwiftUI. Did you make any headway on that?
> 
> Hey! Thanks for taking the time to look it over. For now I left it as is, but I'm looking into doing screenshot tests instead.
> 
> As for writing tests, I don't think we've defined the process for external contributions yet. Let me check with the team to see what the best way to move forward would be

I see, I understand! I imagined you were part of a team working on this. As for iOS Screenshot testing, you most likely already know but the iOSSnapshotTestCase (FBSnapshotTestCase) package is really good for that.

PR #13, opened 2022-01-23
iOS

martyu:
Added more screens, reorganized some files


iosApp/iosApp/App/iOSApp.swift
```diff
@@ -0,0 +1,20 @@
+import SwiftUI
+
+@main
+struct iOSApp: App {
+	@StateObject var viewModel = iOSAppViewModel()
```
andymiller813, line 5: These tabs look huge, no?

andymiller813 (approved):

iosApp/iosApp/App/iOSApp.swift
martyu: Hmm not to me...I think you can set it in your github file viewer?  I use 4 width

PR #14, opened 2022-01-25
KMM POC

andymiller813:
Next up: 
- [x] Include SqlDelight KMM cache example
- [x] Cleanup dependencies files


buildSrc/src/main/java/Ktor.kt
```diff
@@ -0,0 +1,7 @@
+object Ktor {
```
andymiller813, line 1: This is how I ultimately want our dependencies structured.  Deps file was already getting too unwieldy

shared/src/iosMain/kotlin/com/projectnewm/example/di/ExampleNetworkModule.kt
```diff
@@ -0,0 +1,14 @@
+package com.projectnewm.example.di
```
martyu, line 1: I think this file can go in commonMain?

shared/src/commonMain/kotlin/com/projectnewm/BaseUseCase.kt
```diff
@@ -0,0 +1,9 @@
+package com.projectnewm
+
+import com.projectnewm.util.CommonFlow
+import com.projectnewm.util.DataState
+
+interface BaseUseCase<T> {
+
+    fun execute(): CommonFlow<DataState<T>>
```
martyu: Note: generic interfaces don't map to Swift, so this `T` gets converted to Swift's `AnyObject`.

shared/src/commonMain/kotlin/com/projectnewm/example/ExampleUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.projectnewm.util.DataState
+import com.projectnewm.util.asCommonFlow
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.flow
+
+class ExampleUseCase constructor(
+    private val exampleRepository: ExampleRepository
+) : BaseUseCase<List<ExampleDomainModel>> {
+
+    override fun execute(): CommonFlow<DataState<List<ExampleDomainModel>>> =
```
martyu: Note: The Swift version of this loses type information about the List's generic, so it's just an untyped `NSArray` :/

martyu: Wondering if this needs to be a flow or can just suspend and return the model?  Then view models would handle loading state and use cases would be synchronous.

androidApp/src/main/java/com/projectnewm/di/NetworkModule.kt
```diff
@@ -0,0 +1,30 @@
+package com.projectnewm.di
```
martyu, line 1: Should NetworkModule and UseCaseModule have `expect`s?

shared/src/iosMain/kotlin/com/projectnewm/example/di/ExampleNetworkModule.kt
martyu: Or is it not because it has an android version?  In which case should this be an `actual` and have an expect?

martyu (commented):
Pretty slick

martyu (approved):

shared/src/commonMain/kotlin/com/projectnewm/BaseUseCase.kt
andymiller813: Aw bummer! Well, we don't really need it. I'll remove the Base class

shared/src/iosMain/kotlin/com/projectnewm/example/di/ExampleNetworkModule.kt
andymiller813: Seems like we need to keep all the hilt stuff on Android side so we wouldn't have this in shared code for Android

andymiller813: We can revisit if we need another DI solution that works for KMM. I'll need to dig into that some more

shared/src/commonMain/kotlin/com/projectnewm/example/ExampleUseCase.kt
andymiller813: I think it'll be cleaner to have the use case handle loading/data/error emissions and easier to test since everything is in one flow. Plus it's more code that we can share

androidApp/src/main/java/com/projectnewm/di/NetworkModule.kt
andymiller813: See note above about Hilt

andymiller813:
I'll wait for @cristhianescobar  to review before I merge

androidApp/src/main/java/com/projectnewm/di/UseCaseModule.kt
```diff
@@ -0,0 +1,31 @@
...
+    @Provides
+    fun provideExampleUseCase(
+        exampleRepository: ExampleRepository,
+    ): ExampleUseCase {
+        return ExampleUseCase(exampleRepository)
+    }
+
+    @Singleton
+    @Provides
+    fun provideExampleRepository(
```
andymiller813, line 27: Repository should probably be provided by a separate module but I'd like to refactor this anyway once we move everything into feature modules

androidApp/src/main/java/com/projectnewm/NewmAppComposable.kt
```diff
@@ -1,30 +1,29 @@
...
 import androidx.compose.ui.unit.dp
+import androidx.lifecycle.ViewModelStoreOwner
 import androidx.navigation.NavController
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.rememberNavController
+import com.projectnewm.components.NewmRainbowDivider
 import com.projectnewm.navigation.Navigation
 import com.projectnewm.screens.Screen
 import kotlinx.coroutines.flow.collect
-internal const val TAG_BOTTOM_NAVIGATION = "TAG_BOTTOM_NAVIGATION"
```
newmskywalker, line 24: Do the tests still pass without this tag?

androidApp/src/main/java/com/projectnewm/screens/DetailsScreen.kt
```diff
@@ -32,4 +42,18 @@ fun DetailsScreen() {
             fontSize = 25.sp
         )
     }
+
+    if (viewState.titles.isNotEmpty()) {
+        Log.w("DetailsScreen", "Titles: $viewState")
```
newmskywalker, line 47: nit: these side effects should probably run in a LaunchedEffect lambda

newmskywalker (approved):
💪🏽  This is great!

androidApp/src/main/java/com/projectnewm/NewmAppComposable.kt
andymiller813: Oh weird, this must've been a merge conflict. I'll add it back, good catch

buildSrc/src/main/java/JUnit.kt
```diff
@@ -0,0 +1,6 @@
+object JUnit {
+    const val jUnit = "junit:junit:${Versions.jUnit}"
+    const val jUnitRunner = "androidx.test.runner.AndroidJUnitRunner"
+    const val androidxJUnit = "androidx.test.ext:junit:${Versions.androidxJUnit}"
+    const val androidxComposeJUnit = "androidx.compose.ui:ui-test-junit4:1.0.5"
```
andymiller813: note to self: get version from Versions file

shared/src/commonMain/kotlin/com/projectnewm/example/ExampleUseCase.kt
martyu: Yeah, we want to share as much as possible, but I think it's a bad idea to start saying "well, conceptually this *would* go in the view model, but we're not sharing view models, so let's stick it in the use case, purely so we can share it."  If we want to share view models (and the functionality they're intended to contain), we should look more into that.

androidApp/src/main/java/com/projectnewm/screens/DetailsScreen.kt
andymiller813: Going to leave as is because I want to refactor out all the example stuff in a followup

PR #16, opened 2022-01-28
Create ios-unit-tests.yml

martyu:


ios-unit-tests.yml
```diff
@@ -0,0 +1,47 @@
+name: iOS Unit Tests
+
+on:
+  push:
+    branches: [ development ]
+  pull_request:
+    branches: [ development ]
```
newmskywalker: Should we add a `paths` prop to only run this if there are changes to the iOS folder?

ios-unit-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  build:
+    name: Build and Test default scheme using any available iPhone simulator
```
newmskywalker: Is it possible to run unit tests headless?

ios-unit-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+        env:
+          scheme: ${{ 'default' }}
+          platform: ${{ 'iOS Simulator' }}
+        run: |
+          # xcrun xctrace returns via stderr, not the expected stdout (see https://developer.apple.com/forums/thread/663959)
+          device=`xcrun xctrace list devices 2>&1 | grep -oE 'iPhone.*?[^\(]+' | head -1 | awk '{$1=$1;print}'`
+          if [ $scheme = default ]; then scheme=$(cat default); fi
+          if [ "`ls -A | grep -i \\.xcworkspace\$`" ]; then filetype_parameter="workspace" && file_to_build="`ls -A | grep -i \\.xcworkspace\$`"; else filetype_parameter="project" && file_to_build="`ls -A | grep -i \\.xcodeproj\$`"; fi
+          file_to_build=`echo $file_to_build | awk '{$1=$1;print}'`
+          xcodebuild build-for-testing -scheme "$scheme" -"$filetype_parameter" "$file_to_build" -destination "platform=$platform,name=$device"
```
newmskywalker: 😱 
I wonder if this is something that Fastlane could help with?

newmskywalker (commented):
🚀  Thanks for adding this! Just have a few questions

ios-unit-tests.yml
martyu: I think we'd also want to run it if there's changes to the `shared/commonMain` and `shared/iosMain` folders?  And changes to gradle/other project settings could break it too...might be safer to say "any file _except_ `androidApp/*`/`shared/androidMain/*`"?

newmskywalker: Yeah. I've also found it useful to have them run if the workflow file itself changes

ios-unit-tests.yml
martyu: Do we have Fastlane set up?  This was just the default "build xcode" action template that github had.

ios-unit-tests.yml
martyu: Good question...from what I can find, you would have to use a third party tool, or some workarounds.  But AFAIK launching the simulator isn't a big time suck...

.github/workflows/ios-testing.yml
```diff
@@ -0,0 +1,17 @@
...
+  test:
+    name: Testing iOS app
+    runs-on: macOS-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+      - name: Force XCode 13.2
+        run: sudo xcode-select -switch /Applications/Xcode_13.2.app
+      - name: Testing iOS app
+        run: exec ./.github/scripts/test_app.sh
```
newmskywalker, line 30: Nicee!

ios-unit-tests.yml
martyu: @j-mateo updated paths, think those look ok?

ios-unit-tests.yml
martyu: @j-mateo I made it prettier

ios-unit-tests.yml
newmskywalker: We can just look into it later if build times become an issue

ios-unit-tests.yml
newmskywalker: > Do we have Fastlane set up?  This was just the default "build xcode" action template that github had.

Not yet, but I'll hook it up to these workflows soon

newmskywalker (approved):
The clean up looks very nice! It'll allow us to use that script across different workflows

ios-unit-tests.yml
newmskywalker: Just realized that we can `exclude` paths instead of having to include all of these.


```
paths-ignore:
      - "**.md"
      - androidApp
```
etc

newmskywalker: but we can do that in a follow-up pr if you want

andymiller813 (approved):
Lgtm

PR #18, opened 2022-02-05
CU-21n4174 Adding place holder views to Home Tab

cristhianescobar:
Adding placeholder to the home screen to start discussing more about the interactions that user could take. Also, not very clear if switching tabs should trigger a re-rendering of the whole screen or just the section right below it.


newmskywalker:
Task linked: [CU-21n4174 Add placeholder views to Home Tab](https://app.clickup.com/t/21n4174)

PR #22, opened 2022-02-06
Firebase crashlytics and analytics integration

andymiller813:
Note: need to add google services json file back in without secrets before merge


cristhianescobar:
I think we are missing the  `google-services.json` file

newmskywalker:
> Note: need to add google services json file back in without secrets before merge

Google recommends not adding it at all
https://firebase.google.com/docs/projects/learn-more?authuser=0#config-files-objects

> For open source projects, we generally do not recommend including the app's Firebase config file or object in source control because, in most cases, your users should create their own Firebase projects and point their apps to their own Firebase resources (via their own Firebase config file or object).

Should we just have CI inject it at build time and have local copies for development?

androidApp/build.gradle.kts
```diff
@@ -82,12 +86,14 @@ dependencies {
...
     kapt(Hilt.hiltAndroidCompiler)
 
-    testImplementation(JUnit.jUnit)
+    kaptAndroidTest(Hilt.hiltAndroidCompiler)
 
-    kaptAndroidTest (Hilt.hiltAndroidCompiler)
-    androidTestImplementation(Google.Test.composeUiTestJUnit)
-    androidTestImplementation (Google.Test.hiltAndroidTesting)
+    testImplementation(JUnit.jUnit)
 }
```
newmskywalker, line 99: Looks like some of the files are missing the new line at the end

newmskywalker (approved):
Can you add everyone to the firebase project?
I created a project that I've been using for the artist portal that's meant to be for non-prod use. Maybe this new project can be used for prod?

martyu:
@andymiller813 is this just for android-specific code?  Will I need to add it separate for iOS, and if so, how does it work when shared code crashes?  Would I just send the stack trace/exception that the Shared module returns?

andymiller813:
> @andymiller813 is this just for android-specific code? Will I need to add it separate for iOS, and if so, how does it work when shared code crashes? Would I just send the stack trace/exception that the Shared module returns?

@martyu yeah you will need to add a separate integration for iOS.  Crashlytics should just automatically pick up shared code crashes but we will have to see

PR #28, opened 2022-02-16
Pull google_services.json from repo secrets

newmskywalker:


.github/workflows/android-ui-test.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '11'
+      - name: create google_services.json
```
andymiller813: Should we split this out into a separate script since it's nothing to do with ui tests?

andymiller813 (approved):

.github/workflows/android-ui-test.yml
newmskywalker: Do you mean running it as a separate workflow?

andymiller813: Yes, just because this one is named android-ui-test.  Maybe it just needs to be renamed?

newmskywalker: I think we need to run this step for every workflow that builds the android app, because the file won't be persisted across jobs

andymiller813: Sure, I was thinking we could break things like this out into a reusable workflow so we don't need to duplicate the logic and we can keep the main workflows clean 
https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow

newmskywalker: Ahhh I see what you mean. Yeah that's a good idea!
I'll create a reusable workflow once I get it working

PR #29, opened 2022-02-20
Marty/dev/new views

martyu:


iosApp/iosApp/Song Playing/SongPlayingViewModel.swift
```diff
@@ -0,0 +1,122 @@
...
+	
+	func pauseTapped() {
+		playbackState = .paused
+	}
+	
+	func airplayTapped() {
+		
+	}
+	
+	func shareTapped() {
```
newmskywalker, line 74: Do you think it makes sense to create an enum for the different actions and adding a single function to handle them? 

`func onAction(action: SongPlayingAction)` 

and then just have a switch to dispatch the actions internally? This would keep the api for the viewmodel a bit leaner imo

iosApp/iosApp/Song Playing/SongPlayingViewModel.swift
```diff
@@ -0,0 +1,122 @@
...
+		case playing
+		case paused
+	}
+}
+
+private extension Int {
+	var playbackTimeString: String {
+		let formatter = DateComponentsFormatter()
+		formatter.allowedUnits = [.hour, .minute, .second]
+		formatter.unitsStyle = .positional
```
newmskywalker: does the formatter get created every time the playback time changes? If so, is it possible to take it out of the var so it's only created once?

newmskywalker (approved):
lgtm! 💪🏽

iosApp/iosApp/Song Playing/SongPlayingViewModel.swift
martyu: yeah good call 👍

iosApp/iosApp/Song Playing/SongPlayingViewModel.swift
martyu: I generally avoid wrapping function names in an enum, never cared for the pattern just because it seems like you're just swapping method names for enum cases and giving it a namespace, which sounds like what a protocol should do.  If you just want all the actions grouped together so it's easier to grok and more organized, I could make a ViewModelActionHandler protocol, and move these methods to an extension?

PR #30, opened 2022-02-22
Cu 27gypxm i os develop the login page UI marty ulrich

martyu:


newmskywalker:
Task linked: [CU-27gypxm iOS: Develop the login page UI](https://app.clickup.com/t/27gypxm)

shared/src/commonMain/kotlin/io/projectnewm/login/LoginFieldValidator.kt
```diff
@@ -0,0 +1,28 @@
+package io.projectnewm.login
+
+class LoginFieldValidator {
+    public companion object {
+        public fun validate(email: String, password: String): Boolean {
```
andymiller813: No need for the companion object - 

```suggestion
object LoginFieldValidator {
    public fun validate(email: String, password: String): Boolean {
        .....
```

shared/src/commonMain/kotlin/io/projectnewm/login/LoginFieldValidator.kt
```diff
@@ -0,0 +1,28 @@
...
+            return isValidEmail(email) && isValidPassword(password)
+        }
+
+        private fun isValidEmail(email: CharSequence?): Boolean {
```
andymiller813: you switched from non-null String in the `validate` function to nullable CharSequence here and nullable String in `isValidPassword`. Stick with either `CharSequence?` or `String?` all the way through (probably CharSequence? I think)

shared/src/commonMain/kotlin/io/projectnewm/login/LoginFieldValidator.kt
```diff
@@ -0,0 +1,28 @@
+package io.projectnewm.login
+
+class LoginFieldValidator {
```
andymiller813: 👍

shared/src/commonMain/kotlin/io/projectnewm/login/LoginFieldValidator.kt
```diff
@@ -0,0 +1,28 @@
...
+                        "(" +
+                        "\\." +
+                        "[a-zA-Z0-9][a-zA-Z0-9\\-]{0,25}" +
+                        ")+"
+            )
+            return email?.matches(emailAddressRegex) ?: false
+        }
+
+        private fun isValidPassword(password: String?): Boolean {
+            val password = password?.let { it } ?: return false
```
andymiller813: It's cleanest to just do 
```suggestion
           return if (password == null || password.isBlank()) false else true
```

andymiller813 (approved):

PR #31, opened 2022-02-22
Add gitignore for google-services

andymiller813:


androidApp/.gitignore
```diff
@@ -0,0 +1,2 @@
+/build
+google-services.json
```
newmskywalker, line 2: This is already in the main directory's .gitignore. Do we want to remove it from there?

andymiller813: Ah, missed that. I'll close this then

PR #35, opened 2022-02-25
Restructure project

andymiller813:
Large restructure PR to implement the structure mentioned here: https://projectnewmgroup.slack.com/archives/C02T8NE1M08/p1644682503914539

More work to be done especially in the core-theme module and such.  Eventually we will break out more features into separate modules. This pulls all the example KMM code out of the main app


newmskywalker:
mmm interesting. I wonder why the google-services file didn't get created

andymiller813:
> mmm interesting. I wonder why the google-services file didn't get created

Just fixed it I think. Changed up the directories a tad

androidApps/DemoApp/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
cristhianescobar: Don't think we need this file

androidCore/core-theme/src/main/res/values-night/themes.xml
```diff
@@ -0,0 +1,16 @@
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <!-- Base application theme. -->
+    <style name="Theme.Newmmobile" parent="Theme.MaterialComponents.DayNight.DarkActionBar">
```
cristhianescobar: Maybe not as part of this PR, but we should look into define all theming through compose theme.

androidFeatures/feature-example/build.gradle.kts
```diff
@@ -0,0 +1,60 @@
...
+
+android {
+    compileSdk = Versions.compileSdk
+
+    defaultConfig {
+        minSdk = Versions.minSdk
+        targetSdk = Versions.targetSdk
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles("consumer-rules.pro")
```
cristhianescobar: Don't think we need like `consumerProguardFiles("consumer-rules.pro")`

androidCore/core-theme/src/test/java/io/projectnewm/core/theme/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
+package io.projectnewm.core.theme
+
```
cristhianescobar: Can we delete this as well

androidApps/ProjectNewm/build.gradle.kts
```diff
@@ -33,17 +33,15 @@ android {
...
     buildFeatures {
         compose = true
     }
+
     composeOptions {
         kotlinCompilerExtensionVersion = Versions.composeCompiler
     }
 
-    hilt {
-        enableTransformForLocalTests = true
```
cristhianescobar: Did we not need this anymore for the UI test we have in place?

androidApps/DemoApp/build.gradle.kts
```diff
@@ -0,0 +1,77 @@
...
+        targetSdk = Versions.targetSdk
+        versionCode = 1
+        versionName = "1.0"
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    buildTypes {
+        release {
+            isMinifyEnabled = false
+            proguardFiles(
```
cristhianescobar: We can remove this as well

cristhianescobar (approved):
💯  Overall, it looks good to me, another thing I would suggest is that now that we have a demo android app and the projectnewm app we can probably split out the work related to the kmm POC 

`Proguard`
I think we can go ahead and remove all Proguard files and references, there isn't  a need for it because of the open source nature of our project. 

`theming`
With the core theme module in place we should be able to get rid of a lot of resource files we used to have, with compose we can bring in all the resources through the theme.

cristhianescobar:
After pulling down the branch and looking at the file structure on Android Studio, I think I would be in favor of nesting `androidApps` `androidCore` `androidFeatures` under one `android`  directory to avoid having unnecessary directories at the top level. Just like, there is one iosApp directory.
I think since we will be sharing  a lot more KMM libraries, I suggest we make those top level. Example, we can have  `feature-api-modules` modules at the top level since they are relevant to both OSs. Same for Core libraries that shared across KMM libraries. Any sort of business logic modules would be good candidates for top level directories.

andymiller813:
> 💯 Overall, it looks good to me, another thing I would suggest is that now that we have a demo android app and the projectnewm app we can probably split out the work related to the kmm POC
> 
> `Proguard` I think we can go ahead and remove all Proguard files and references, there isn't a need for it because of the open source nature of our project.
> 
> `theming` With the core theme module in place we should be able to get rid of a lot of resource files we used to have, with compose we can bring in all the resources through the theme.

I split out the example KMM POC stuff into feature-example already and it's only pulled in by the demo app.  Or was there something else you were referring to? I still need to break it out of the main kmm module for sure.

We are still going to want proguard for shrinking and optimizations, we can just disable the obfuscation piece.

👍 Agreed on theming, we will continue to do more cleanup in future PRs

androidApps/ProjectNewm/build.gradle.kts
andymiller813: I got a build warning telling me to remove it 🤷

newmskywalker (approved):

PR #44, opened 2022-03-07
CU-2cr79pa Now playing screen start, break out core-theme

andymiller813:


newmskywalker:
Task linked: [CU-2cr79pa Android: Develop the "Now Playing" screen](https://app.clickup.com/t/2cr79pa)

newmskywalker (approved):
💪🏽

android/app-project-newm/src/main/java/io/projectnewm/screens/Screen.kt
```diff
@@ -7,6 +7,7 @@ sealed class Screen(val route: String) {
     object TribeRoot : Screen("tribe-root")
     object StarsRoot : Screen("stars-root")
     object WalletRoot : Screen("wallet-root")
+    object NowPlayingRoot : Screen("now-playing-root")
```
cristhianescobar, line 10: Do we need the NowPlaying as a root as well as the NowPlayingLanding?

android/feature-now-playing/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+    implementation(project(Modules.coreTheme))
+    implementation(project(Modules.shared))
+
+    implementation(Airbnb.showkase)
+    implementation(Google.androidxCore)
+    implementation(Google.appCompat)
+    implementation(Google.composeMaterial)
+    implementation(Google.composeUi)
+    implementation(Google.composeUiToolingPreview)
+    implementation(Google.constraintLayout)
```
cristhianescobar, line 52: Are we using the constraintLayout somewhere?

cristhianescobar (approved):
💯 Looks good! Great callout on making the now playing its own feature module.

android/app-project-newm/src/main/java/io/projectnewm/screens/Screen.kt
andymiller813: Honestly not sure I was just following the pattern there :D I can play around with it in a follow up. I was wondering the same

android/feature-now-playing/build.gradle.kts
andymiller813: This may be an artifact of the pre-jetpack days and copy pasta here.  I want to go through and create some gradle scripts that make it so we don't have to copy paste every dependency everywhere, so I'll clean this up later.

PR #54, opened 2022-03-16
CU-2e0d390 Hide or show bottom nav bar on a per-screen basis

andymiller813:


newmskywalker:
Task linked: [CU-2e0d390 Create a way to hide the bottom nav on a per-screen basis](https://app.clickup.com/t/2e0d390)

android/app-project-newm/src/main/java/io/projectnewm/NewmAppComposable.kt
```diff
@@ -80,52 +95,60 @@ private fun DebugMenuButton() {
...
+            }
         }
     }
 }
 
 @Preview(showBackground = true)
 @Composable
 fun BottomNavigationBarPreview() {
-    NewmBottomNavigation(Screen.HomeRoot) {}
+    NewmBottomNavigation(Screen.HomeRoot, true) {}
```
cristhianescobar, line 152: Is there a different approach we can take? When we had the detail page example it took the entire screen, can do something similar for the now playing?

I don't see the case where we would need to show the bottom navigation if we are not in the main screens. And we will need to pass in false for any subsequent screens, which there will probably be more than the numbers of screens we have on our bottom nav

andymiller813: This is the best way I could figure out for now with compose.  We can address it later once we get more screens but for now  it's straightforward enough to just add the  route to the `routesWithoutBottomNavBar` list and the rest will happen automatically

andymiller813: If you have another suggestion I'm game to try it but otherwise let's keep this as-is for now since it works great for the demo

cristhianescobar (approved):
👍🏼 Make sense to keep it as it is for demo purposes, maybe we can revisit the approach once the complexity increases

PR #55, opened 2022-03-19
Cu 2g3hp2d i os split features into swift packages marty ulrich

martyu:
Split app into feature modules.


newmskywalker:
Task linked: [CU-2g3hp2d iOS: Split features into modules](https://app.clickup.com/t/2g3hp2d)

andymiller813:
How come you put TabBar and PlaylistList in their own modules?

martyu:
> How come you put TabBar and PlaylistList in their own modules?

TabBar because it's an isolated UI component and I didn't see a reason not to 🤷 do you?

PlaylistList because it's a screen, just like Home/SongPlaying/Tips.  Where would you expect to see it?

andymiller813 (approved):

andymiller813:
I recommend leaving the tab bar in main until we know it needs to be shared, and then maybe adding it to Shared UI.  
I also recommend prefixing features with `feature-` or `Feature` and core modules with `core-` or `Core` so at first glance you know how each module is being used.

martyu:
Split modules into Screens, Resources, and Helpers folders

PR #59, opened 2022-04-03
CU-2arvjy2 Wiring Login API through KMM shared Module

cristhianescobar:
Adding share service work for login API

Perhaps we can keep in the share module for now


newmskywalker:
Task linked: [CU-2arvjy2 Wiring Login API through KMM shared Module](https://app.clickup.com/t/2arvjy2)

shared/src/commonMain/kotlin/io.projectnewm.shared/login/Users.kt
```diff
@@ -0,0 +1,13 @@
+package io.projectnewm.shared.login
+
+data class LoginUser(val email: String, val password: String)
```
martyu: nit: login is a noun, log in is an action

```suggestion
data class LogInUser(val email: String, val password: String)
```

martyu (approved):

martyu (changes requested):
sorry didn't mean to approve, still reviewing

android/app-project-newm/src/main/java/io/projectnewm/di/usecase/LoginModule.kt
```diff
@@ -0,0 +1,46 @@
...
+
+@Module
+@InstallIn(SingletonComponent::class)
+object LoginModule {
+
+    @Singleton
+    @Provides
+    fun provideLoginUseCase(
+        repository: LoginRepository,
+    ): LoginUseCase {
```
martyu: Where would the public function be that clients call?  LoginRepository wouldn't be public, so would there be a public function with signature `public fun foo() -> LogInUseCase`?

shared/src/commonMain/kotlin/io.projectnewm.shared/login/LoginRepository.kt
```diff
@@ -0,0 +1,40 @@
+package io.projectnewm.shared.login
+
+//TODO: Handle Error Cases
+class LoginRepository(
+    private val service: LoginService
+) {
+    suspend fun registerUser(email: String): String {
+        return service.requestSignUpAccess(email)
```
martyu: What is the string being returned here?  Also would have expected this function to also be called `requestEmailConfirmationCode` or something.

shared/src/commonMain/kotlin/io.projectnewm.shared/login/LoginService.kt
```diff
@@ -0,0 +1,38 @@
...
+import io.ktor.client.*
+import io.ktor.client.request.*
+
+interface LoginService {
+    suspend fun requestSignUpAccess(email: String): String
+    suspend fun register(user: NewUser): String
+    suspend fun login(user: LoginUser): String
+
+    companion object {
+        const val BASE_URL = "https://staging-newm-server.herokuapp.com"
```
martyu: Do we want this URL in a common location for reuse?

shared/src/commonMain/kotlin/io.projectnewm.shared/login/LoginService.kt
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+
+    override suspend fun register(user: NewUser): String {
+        return httpClient.put(body = user) {
+            url("$baseUrl/v1/users")
+        }
+    }
+
+    override suspend fun login(user: LoginUser): String {
```
martyu: nit: 
```suggestion
    override suspend fun logIn(user: LoginUser): String {
```

martyu (dismissed):

shared/src/commonMain/kotlin/io.projectnewm.shared/login/LoginService.kt
cristhianescobar: We can find a better home for this once we implement other services and see what makes sense to refactor. 
Rule of three

android/app-project-newm/src/main/java/io/projectnewm/di/usecase/LoginModule.kt
cristhianescobar: Yes, we can consume the LogInRepositroy from a use case or two, not sure if we want to make multiple use cases to address user sign up and user log In.

cristhianescobar: At this moment the `LoginRepository` is public and doing the initialization via DI.  But I can also see us keeping the repository internal to the library if we figure out a DI solution for the common modules that work well with both platforms

shared/src/commonMain/kotlin/io.projectnewm.shared/login/LoginRepository.kt
cristhianescobar: The string being returned here would be a token, will update it once we have a usecase to test out the full integration with the UI

andymiller813 (approved):

shared/src/commonMain/kotlin/io.projectnewm.shared/login/LoginFeature.kt
```diff
@@ -0,0 +1,12 @@
+package io.projectnewm.shared.login
+
+import io.ktor.client.*
+import io.projectnewm.shared.KtorClientFactory
+
+object LoginFeature {
```
cristhianescobar: While we figure our DI tooling, we can use this class to test out the endpoints.

martyu:
Are we going to be exposing `LoginFeature` publicly, or are you saying that's just temporary until we get DI set up?  I would have expected the only thing that was publicly exposed to the client apps was `LogInUseCase`.

cristhianescobar:
> Are we going to be exposing `LoginFeature` publicly, or are you saying that's just temporary until we get DI set up? I would have expected the only thing that was publicly exposed to the client apps was `LogInUseCase`.

I was thinking about  exposing`LoginFeature` for now until we find a solid DI solution

andymiller813:
@martyu @cristhianescobar Are we waiting for anything else on this PR?

PR #61, opened 2022-04-26
CU-284cuwe Basic audio streaming of adaptive bitrate files from s3

andymiller813:


newmskywalker:
Task linked: [CU-2e0d3av Clean up gradle scripts by creating helper scripts](https://app.clickup.com/t/2e0d3av)

android/feature-now-playing/src/main/java/io/projectnewm/feature/now/playing/NowPlayingScreen.kt
```diff
@@ -70,6 +91,12 @@ fun NowPlayingScreen() {
         Spacer(modifier = Modifier.height(24.dp))
 
         TipIcon()
+
+        if (firstLoad.value) {
+            Player.play(context)
+            firstLoad.value = false
+            isPlaying.value = true
+        }
```
newmskywalker: Could `LaunchedEffect(Unit) { Player.play(context) }` potentially replace the `firstLoad` state?

newmskywalker (approved):
Looks good! Just a minor comment about potentially using `LaunchedEffect`

android/feature-now-playing/src/main/java/io/projectnewm/feature/now/playing/NowPlayingScreen.kt
andymiller813: Updated, I think that works

PR #62, opened 2022-05-26
Replace hilt with Koin

cristhianescobar:
As I was looking through DI tooling, realized that hilt is only helpful within Android, for anything KMM related hilt wasn't very helpful so after doing some digging and talking to a few people I came to the conclusion that Koin might be a good candidate as it will help us in the KMM space


android/app-demo/src/main/java/io/projectnewm/demo/di/CacheModule.kt
```diff
@@ -1,28 +1,9 @@
...
-    @Singleton
-    @Provides
-    fun provideDao(
-        newmDatabase: NewmDb
-    ): ExampleDao {
-        return ExampleDaoImpl(db = newmDatabase)
-    }
-}
\ No newline at end of file
+val cacheModule = module {
```
andymiller813, line 6: 0 annotations! Love it already lol

andymiller813 (commented):
Did you try some DI within kmm yet?

buildSrc/src/main/java/Koin.kt
```diff
@@ -0,0 +1,15 @@
+object Koin {
+    private const val VERSION = "3.2.0"
+
+    const val core = "io.insert-koin:koin-core:${VERSION}"
+    const val ktor = "io.insert-koin:koin-ktor:${VERSION}"
+
+    const val android = "io.insert-koin:koin-android:${VERSION}"
+    const val androidViewModel = "io.insert-koin:koin-androidx-viewmodel:2.2.3"
```
wlara: Should this should be `${VERSION}` rather than hard-coded to `2.2.3` ?

android/app-project-newm/src/main/java/io/projectnewm/di/usecase/LoginModule.kt
```diff
@@ -1,21 +1,8 @@
...
-object LoginModule {
-
-    @Singleton
-    @Provides
-    fun provideLoginUseCase(): LoginUseCase {
-        return LoginConfig.getLoginUseCase()
-    }
+import org.koin.dsl.module
 
+val login = module {
```
wlara: Rename to `loginModule` for consistency with other module names

buildSrc/src/main/java/Koin.kt
cristhianescobar: Koin Android View Model has a different version code than the rest of Koin, but I can make this version a variable

cristhianescobar:
> Did you try some DI within kmm yet?

I have, it's coming on my next PR

wlara (approved):

PR #65, opened 2022-06-14
Wire Signup and Login Use cases

cristhianescobar:


android/app-project-newm/src/main/java/io/projectnewm/MainActivity.kt
```diff
@@ -25,16 +26,45 @@ class MainActivity : ComponentActivity() {
 }
 
 @Composable
-fun WelcomeToNewm() {
+fun WelcomeToNewm(signupViewModel: SignupViewModel = org.koin.androidx.compose.get()) {
```
wlara, line 29: what do you think about using `koinViewModel()` here, like:
```
fun WelcomeToNewm(signupViewModel: SignupViewModel = koinViewModel())
```

android/feature-login/src/main/java/io/projectnewm/feature/login/screen/LoginScreen.kt
```diff
@@ -36,29 +35,24 @@ internal const val TAG_LOGIN_SCREEN = "TAG_LOGIN_SCREEN"
 
 @Composable
 fun LoginScreen(
-    onSignInSubmitted: (email: String, password: String) -> Unit,
-    scrollState: ScrollState = rememberScrollState()
+    onUserLoggedIn: () -> Unit,
+    onSignupClick: () -> Unit,
+    viewModel: LoginViewModel = org.koin.androidx.compose.get()
```
wlara, line 40: Same here. Maybe use `koinViewModel()`.

wlara (approved):

martyu:
@cristhianescobar is this building for you?  I get "Unresolved reference" errors for LoginStatus and isValid.

shared/src/commonMain/kotlin/io.projectnewm.shared/login/models/LoginResponse.kt
```diff
@@ -0,0 +1,35 @@
+package io.projectnewm.shared.login.models
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class LoginResponse(
```
martyu, line 7: I don't think LoginResponse needs to be returned by LoginUseCase does it?  I thought accessToken and refreshToken would be stored in KMM.  I don't think the clients would need to access them directly eh?

shared/src/commonMain/kotlin/io.projectnewm.shared/login/models/LoginResponse.kt
```diff
@@ -0,0 +1,35 @@
...
+    @SerialName("accessToken")
+    val accessToken: String,
+    @SerialName("refreshToken")
+    val refreshToken: String
+)
+
+sealed class LoginStatus {
```
martyu, line 14: Is there a way we can just throw these errors?  Or if we decide to share view models, we can just set the error there and the client view's can show them.

martyu (approved):

PR #68, opened 2022-08-21
Implement home screen v1

martyu:
This is massive.


cristhianescobar (approved):
Let's merge this, but try not to have such big PRs going forward 😂

PR #70, opened 2022-08-28
Update library versions

cristhianescobar:


iosApp/iosApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,32 @@
+{
```
cristhianescobar, line 1: @martyu Every time I open AppCode to ensure I didn't break anything on the iOS app I get this file updated, so I added it for now? If it causes any issues on your side I can remove it through .gitignore

martyu: Not sure why I don't see it, but yeah should be fine

martyu (approved):

PR #76, opened 2022-10-26
KMM Beta Update

cristhianescobar:
Update project to use the latest kmm beta release as part of this work, I also had to do the following:
-Removed unused code
-Update ktor
-Update newmAPI implementation
-Reworked DI, so we have native and common hooks to add stuff to the DI graph 
-Rework DB setup


wlara (approved):

AndrewWestberg (approved):
Not an expert in this, but approving in case it's blocking you.

PR #82, opened 2022-11-18
Music player

martyu:
- Add AudioPlayer stub as a temporary solution for clickthru mvp (to be moved to KMM).
- Add SwiftGen to generate swift code for asset catalogs.


cristhianescobar (approved):
LGTM

PR #109, opened 2023-04-19
Login/Marketplace

martyu:
Lots of shit :)


shared/src/commonMain/kotlin/io.newm.shared/di/Koin.kt
```diff
@@ -79,4 +86,31 @@ class LoginUseCaseFactory() : KoinComponent {
     fun loginUseCase(): LoginUseCase {
         return loginUseCase
     }
+}
+
+class SignupUseCaseFactory() : KoinComponent {
```
cristhianescobar, line 91: Seems like we will have a lot of repetitions with all these factories... What if we create a `ClientCaseFactory` and provide all the cases there? Maybe we can move that to a separate file.

shared/src/commonMain/kotlin/io.newm.shared/playlist/repository/GenresRepository.kt
```diff
@@ -0,0 +1,21 @@
...
+    suspend fun getGenres(): List<String>
+}
+
+internal class GenresRepositoryImpl : KoinComponent, GenresRepository {
+    private val service: GenresAPI by inject()
+
+    private val logger = Logger.withTag("NewmKMM-GenreRepo")
+
+    override suspend fun getGenres(): List<String> {
+        logger.d { "getGenres" }
```
cristhianescobar, line 18: It might be a good idea to return a flow from the database, and once the response is successful, we can store the network response into the DB.

shared/src/commonMain/kotlin/io.newm.shared/playlist/usecases/GetGenresUseCase.kt
```diff
@@ -0,0 +1,18 @@
...
+
+interface GetGenresUseCase {
+    @Throws(KMMException::class, CancellationException::class)
+    suspend fun getGenres(): List<String>
+}
+
+internal class GetGenresUseCaseImpl(private val repository: GenresRepository) : GetGenresUseCase {
+
+    @Throws(KMMException::class, CancellationException::class)
+    override suspend fun getGenres(): List<String> {
```
cristhianescobar, line 15: Eventually all use cases should return flowables

cristhianescobar (approved):

PR #124, opened 2023-06-10
CU-866a8gzby | CU-866a8gzd1 | CU-866a8gzm7 - Implemented Profile Screen Text Fields

Jermainelr:
Screenshot_20230610_134826
Screenshot_20230610_134833


newmskywalker:
Task linked: [CU-866a8gzd1 Implement "Change Password" Text Fields](https://app.clickup.com/t/866a8gzd1)

newmskywalker:
Task linked: [CU-866a8gzby Implement "Your Profile" Text Fields](https://app.clickup.com/t/866a8gzby)

newmskywalker:
Task linked: [CU-866a8gzm7 Integrate "Save Button" Component into UI](https://app.clickup.com/t/866a8gzm7)

android/app-newm/src/main/java/io/newm/screens/profile/ProfileForm.kt
```diff
@@ -0,0 +1,125 @@
...
+    fontWeight = FontWeight.Bold,
+    color = Gray100
+)
+
+private val formTextFieldStyle = TextStyle(
+    fontSize = 16.sp,
+    fontFamily = inter,
+    fontWeight = FontWeight.Normal,
+    color = Gray100
+)
```
newmskywalker, line 43: Does it make sense to put these styles in the theme and pull them from there instead?

android/app-newm/src/main/java/io/newm/screens/profile/ProfileForm.kt
```diff
@@ -0,0 +1,125 @@
...
+
+@Composable
+fun ProfileForm(
+    profile: ProfileModel,
```
newmskywalker, line 47: A potential future improvement would be to create a different composable to extract the state from the viewmodel. That way you can decouple the view from the viewmodel and test it in isolation (check out `LoginScreen` for an example

android/app-newm/src/main/java/io/newm/screens/profile/ProfileForm.kt
```diff
@@ -0,0 +1,125 @@
...
+        visualTransformation = if (isPassword) PasswordVisualTransformation() else VisualTransformation.None,
+        textStyle = formTextFieldStyle,
+        colors = TextFieldDefaults.textFieldColors(
+            focusedIndicatorColor = Color.Transparent,
+            unfocusedIndicatorColor = Color.Transparent,
+        ),
+        modifier = Modifier
+            .fillMaxWidth()
+            .border(width = 2.dp, color = Gray500)
+    )
```
newmskywalker, line 124: Should we put this in a common module so it can be used in other screens?

android/core-theme/src/main/java/io/newm/core/theme/Color.kt
```diff
@@ -10,7 +10,7 @@ val White = Color.White
 val Black = Color.Black
 val Purple = Color(0xFFDC3CAA)
 val Pinkish = Color(0xFFF53C69)
-val Gray = Color(0xFF1C1C1E)
+val Gray500 = Color(0xFF1C1C1E)
```
newmskywalker, line 13: @cristhianescobar we should look into how we want to consume colors. Ideally, we'd want to use theme values so we can update colors without having to refactor the whole app

newmskywalker (approved):
Looks good! 
Just some comments about reusability

android/app-newm/src/main/java/io/newm/screens/profile/ProfileForm.kt
cristhianescobar: @Jermainelr let's move this element to the `core-ui-utils`, we will probably end up renaming this module in the near future to make it explicit that these modules holds our custom UI components

android/core-theme/src/main/java/io/newm/core/theme/Color.kt
cristhianescobar: I like that idea, we can probably do that outside the scope of this PR. But that is something we need to tackle. Perhaps we can start using github issues to track what needs to be done for now.

PR #127, opened 2023-07-11
CU-862k43hfa - Add circuit and decouple viewmodel from view

newmskywalker:


newmskywalker:
Task linked: [CU-862k43hfa Add circuit and decouple viewmodel from view](https://app.clickup.com/t/862k43hfa)

android/feature-login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountUiState.kt
```diff
@@ -0,0 +1,14 @@
+package io.newm.feature.login.screen.createaccount
+
+import com.slack.circuit.runtime.CircuitUiState
+import io.newm.feature.login.screen.TextFieldState
+import io.newm.feature.login.screen.createaccount.signupform.SignupFormUiEvent
+
+sealed interface CreateAccountUiState : CircuitUiState {
+    data class SignupForm(
```
cristhianescobar, line 8: This looks great, curious with this approach will we be able to reuse states like Loading, Error or will those states have to exist within CreateAccountUiState seal class?

cristhianescobar (approved):

android/feature-login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountPresenter.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.compose.runtime.remember
+import com.slack.circuit.retained.rememberRetained
+import com.slack.circuit.runtime.Navigator
+import com.slack.circuit.runtime.presenter.Presenter
+import io.newm.feature.login.screen.TextFieldState
+import io.newm.feature.login.screen.createaccount.signupform.SignupFormUiEvent
+import io.newm.feature.login.screen.password.ConfirmPasswordState
+import io.newm.feature.login.screen.password.PasswordState
+
+class CreateAccountPresenter : Presenter<CreateAccountUiState> {
```
wlara, line 13: Never heard about this Circuit library before. They should have used a different name instead of `Presenter`, because of all the old MVP architecture baggage. Or is Circuit really Mvp?

newmskywalker: It's MVP with unidirectional data flow

android/feature-login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountUiState.kt
newmskywalker: We could technically reuse them, but the overhead is not worth it

PR #128, opened 2023-07-18
User/me API Integration

cristhianescobar:
Allow logged-in users to see their personal data in the profile page.
Allow logged-in users to log out.


shared/src/iosMain/kotlin/shared/SecureStorage.kt
```diff
@@ -0,0 +1,63 @@
...
+import platform.Security.kSecAttrService
+import platform.Security.kSecClass
+import platform.Security.kSecClassGenericPassword
+import platform.Security.kSecMatchLimit
+import platform.Security.kSecMatchLimitOne
+import platform.Security.kSecReturnData
+import platform.Security.kSecValueData
+import platform.darwin.errSecSuccess
+
+actual class SecureStorage(private val account: String) {
```
cristhianescobar, line 20: @martyu before consuming the login api we might need to revisit this area.

android/app-newm/src/main/java/io/newm/Logout.kt
```diff
@@ -0,0 +1,37 @@
...
+    private val userSession: UserSession,
+    private val restartApp: RestartApp,
+) : KoinComponent, Callable<Int> {
+
+    private val logger = Logger.withTag("NewmAndroid-Logout")
+
+    override fun call(): Int {
+        return try {
+            userSession.logout()
+            restartApp.run()
```
newmskywalker, line 20: Why do we need to restart the app when the user logs out?

newmskywalker (approved):

android/app-newm/src/main/java/io/newm/Logout.kt
cristhianescobar: We do that to make sure anything in memory goes away, typical approach I would say.

PR #137, opened 2023-09-13
Restructure modules to separate features from core libraries

cristhianescobar:
Going forward there will be a place for core libraries to live, separated from stand alone features
Screenshot 2023-09-13 at 12 15 08 AM


buildSrc/src/main/java/Modules.kt
```diff
@@ -1,10 +1,10 @@
...
-    const val coreUiUtils = ":android:core-ui-utils"
-    const val coreResources = ":android:core-resources"
+    const val coreTheme = ":android:core:theme"
+    const val coreUiUtils = ":android:core:ui-utils"
+    const val coreResources = ":android:core:resources"
 
-    const val nowPlaying = ":android:feature-now-playing"
-    const val login = ":android:feature-login"
+    const val nowPlaying = ":android:features:now-playing"
+    const val login = ":android:features:login"
```
wlara, line 7: Nice! This is cleaner and consistent with the latest MAD trend.

wlara (approved):

PR #145, opened 2023-10-13
Barcode Scanner Activity

cristhianescobar:
Adding a way for qr codes to be scanned, the activity gets called and the calling activity will get a string once there is a xpub key scanned.


Jermainelr (approved):
LGTM! Using Google ML Kit for the Barcode Activity looks like a clean implementation

cristhianescobar:
> LGTM! Using Google ML Kit for the Barcode Activity looks like a clean implementation

We can probably clean out the implementation a bit more, once we get the finalized designs we can do some of it.

PR #150, opened 2023-10-30
Implement Wallet Connect UseCase

cristhianescobar:
Adding a use case for wallet connection:

- Remove unnecessary interfaces from the repository layer, as that layer is internal to KMM.
- The only thing for client consumption from now on is the use cases.


newmskywalker (approved):

shared/src/commonMain/kotlin/io.newm.shared/repositories/WalletConnectManager.kt
```diff
@@ -0,0 +1,28 @@
...
+        storage.deleteObject(WALLET_CONNECT_XPUB_KEY)
+    }
+
+    fun isConnected(): Boolean {
+        return storage.string(WALLET_CONNECT_XPUB_KEY) != null
+    }
+
+    fun getXpub(): String? {
+        return storage.string(WALLET_CONNECT_XPUB_KEY)
+    }
```
martyu, line 23: Should `getXpub` be internal?

shared/src/commonMain/kotlin/io.newm.shared/repositories/WalletConnectManager.kt
```diff
@@ -0,0 +1,28 @@
...
+        storage.set(WALLET_CONNECT_XPUB_KEY, xpub)
+    }
+
+    fun disconnect() {
+        storage.deleteObject(WALLET_CONNECT_XPUB_KEY)
+    }
+
+    fun isConnected(): Boolean {
+        return storage.string(WALLET_CONNECT_XPUB_KEY) != null
+    }
```
martyu, line 19: `    fun isConnected(): Boolean {
        return getXpub() != null
    }`

PR #154, opened 2023-11-07
Initial wallet data parsing

cristhianescobar:
Initial parsing of V1 and V2 version of the metadata

Screenshot 2023-11-07 at 12 15 06 AM


android/features/music-player/src/main/java/io/newm/feature/musicplayer/repository/MusicRepository.kt
```diff
@@ -4,9 +4,9 @@ import android.content.Context
 import android.net.Uri
 import io.newm.feature.musicplayer.models.Playlist
 import io.newm.feature.musicplayer.models.Track
+import io.newm.feature.musicplayer.R
```
martyu, line 7: what's `R`?

android/app-newm/src/main/java/io/newm/navigation/Navigation.kt
```diff
@@ -17,20 +17,19 @@ import io.newm.screens.library.LibraryScreen
 import io.newm.screens.library.NFTLibraryScreen
 import io.newm.screens.profile.edit.ProfileRoute
 import io.newm.screens.profile.view.ProfileViewScreen
-import io.newm.shared.models.Song
+import io.newm.shared.repositories.NFTTrack
```
martyu: repositories shouldn't be exposed to the Presentation layer right?  The whole data layer (API/DAO) should be internal to the KMM module.

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryViewModel.kt
```diff
@@ -1,56 +1,48 @@
...
-import kotlinx.coroutines.flow.flatMapLatest
-import kotlinx.coroutines.flow.mapLatest
-import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.launch
 
 class NFTLibraryViewModel(
-    private val useCase: WalletNFTSongsUseCase,
-    private val walletConnectManager: WalletConnectUseCase
+    private val walletConnectUseCase: WalletConnectUseCase,
```
martyu: We shouldn't use the term "WalletConnect" since we might add the actual "WalletConnect" protocol at some point.  "ConnectWallet" would be less ambiguous.

shared/src/commonMain/kotlin/io.newm.shared/repositories/CardanoWalletRepository.kt
```diff
@@ -33,15 +34,10 @@ fun List<LedgerAssetMetadata>.getMusicMetadataVersion(): Int? {
...
-    val files: List<File>,
-    val musicMetadataVersion: Int,
-    val releaseType: String?,
-    val releaseTitle: String?,
-    val distributor: String?,
-    val publicationDate: String?
+    val songName: String,
+    val imageUrl: String,
+    val songSrc: String,
+    val artist: List<String> = emptyList(),
```
martyu:
```suggestion
    val artists: List<String> = emptyList(),
```

shared/src/commonMain/kotlin/io.newm.shared/repositories/CardanoWalletRepository.kt
```diff
@@ -33,15 +34,10 @@ fun List<LedgerAssetMetadata>.getMusicMetadataVersion(): Int? {
...
-    val imageMimeType: String,
-    val files: List<File>,
-    val musicMetadataVersion: Int,
-    val releaseType: String?,
-    val releaseTitle: String?,
-    val distributor: String?,
-    val publicationDate: String?
+    val songName: String,
+    val imageUrl: String,
+    val songSrc: String,
```
martyu:
```suggestion
    val songSrc: URL,
```

shared/src/commonMain/kotlin/io.newm.shared/repositories/CardanoWalletRepository.kt
```diff
@@ -33,15 +34,10 @@ fun List<LedgerAssetMetadata>.getMusicMetadataVersion(): Int? {
...
-    val image: String,
-    val imageMimeType: String,
-    val files: List<File>,
-    val musicMetadataVersion: Int,
-    val releaseType: String?,
-    val releaseTitle: String?,
-    val distributor: String?,
-    val publicationDate: String?
+    val songName: String,
+    val imageUrl: String,
```
martyu:
```suggestion
    val imageUrl: URL,
```

shared/src/commonMain/kotlin/io.newm.shared/repositories/CardanoWalletRepository.kt
```diff
@@ -33,15 +34,10 @@ fun List<LedgerAssetMetadata>.getMusicMetadataVersion(): Int? {
...
-    val name: String,
-    val image: String,
-    val imageMimeType: String,
-    val files: List<File>,
-    val musicMetadataVersion: Int,
-    val releaseType: String?,
-    val releaseTitle: String?,
-    val distributor: String?,
-    val publicationDate: String?
+    val songName: String,
```
martyu: Should this be `trackName` to be consistent with the class type?  In which case `name` would be fine to avoid redundency.

shared/src/commonMain/kotlin/io.newm.shared/repositories/CardanoWalletRepository.kt
```diff
@@ -50,38 +46,47 @@ data class File(
...
+        ?.flatMap { artist ->
+            artist.children.mapNotNull { detail ->
+                if (detail.key == "name") detail.value as? String else null
+            }
+        }.orEmpty()
+
+    val songSrcId = this.find { it.key == "image" }?.value ?: return null
+
+    return NFTTrack(
+        songName = find { it.key == "song_title" }?.value ?: "Where is the song name?",
```
martyu: Doesn't this mean "Where is the song name?" will be shown to users?  If it's not a required field then it should be optional.  But I would expect it to be required, and if it's missing then the return value should be nil, or throw an error.  I would expect it to be required and throw an error, if it's not possible to mint a song with no title.

shared/src/commonMain/kotlin/io.newm.shared/repositories/CardanoWalletRepository.kt
```diff
@@ -50,38 +46,47 @@ data class File(
...
-        imageMimeType,
-        fileList,
-        musicMetadataVersion,
-        releaseType,
-        releaseTitle,
-        distributor,
-        publicationDate
+        songName = name,
+        imageUrl = "https://arweave.net/${image.replace("ar://", "")}",
+        songSrc = "https://bcsh.mypinata.cloud/ipfs/${source.replace("ar://", "")}",
```
martyu: Aren't IPFS and Arweave different?

martyu (changes requested):

shared/src/commonMain/kotlin/io.newm.shared/usecases/WalletNFTSongsUseCase.kt
```diff
@@ -10,7 +10,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlin.coroutines.cancellation.CancellationException
 
 interface WalletNFTSongsUseCase {
-    fun getAllWalletNFTSongs(): Flow<List<Song>>
+    fun getAllWalletNFTSongs(): Flow<List<NFTTrack>>
```
martyu: What was this for again?  Flows can't be used on iOS.

shared/src/commonMain/kotlin/io.newm.shared/repositories/CardanoWalletRepository.kt
martyu: Are these the artist's display names or identifiers?  Thinking maybe they should be identifiers, since (eventually) we'll want to be able to let the user tap the artist name and open the artist page, like spotify.

android/features/music-player/src/main/java/io/newm/feature/musicplayer/repository/MusicRepository.kt
cristhianescobar: This is Android specific, it's an import required when referencing resources.

android/app-newm/src/main/java/io/newm/navigation/Navigation.kt
cristhianescobar: That is correct, all classes in KMM will either be internal or private except for use cases and data models.

martyu (approved):

PR #156, opened 2023-11-15
NFT metadata parsing version 1 & 2

cristhianescobar:


shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
```diff
@@ -34,59 +35,197 @@ fun List<LedgerAssetMetadata>.getMusicMetadataVersion(): Int? {
...
-            artist.children.mapNotNull { detail ->
-                if (detail.key == "name") detail.value else null
+            "files" -> {
+                metadata.children.forEach { files ->
+                    val isThereAudioFiles = files.children.find { file ->
+                        file.key == "mediaType" && file.value.contains("audio")
+                    } != null
+
+                    if (isThereAudioFiles) {
+                        files.children.forEach { file ->
```
martyu: should this be a "first" instead of "forEach" so we don't keep iterating once the file is found?

shared/src/commonMain/kotlin/io.newm.shared/public/models/NFTTrack.kt
```diff
@@ -2,8 +2,10 @@ package io.newm.shared.public.models
 
 
 data class NFTTrack(
+    val id: String,
     val name: String,
     val imageUrl: String,
     val songUrl: String,
+    val duration: String,
```
martyu: Either `Int` or `Float`/`Double`

martyu (approved):

PR #158, opened 2023-11-18
Enhancement of Code Documentation with KDoc

cristhianescobar:
Added KDoc to key interfaces and data models visible to mobile clients. Also rearranged Gradle dependencies for better clarity.


shared/src/commonMain/kotlin/io.newm.shared/public/usecases/LoginUseCase.kt
```diff
@@ -1,30 +1,73 @@
...
+     * Authenticates a user using their Apple ID with the provided ID token.
+     *
+     * @param idToken The ID token from Apple Sign-In.
+     * @throws KMMException if there is an issue during the login process.
+     * @throws CancellationException if the login process is cancelled.
+     */
     @Throws(KMMException::class, CancellationException::class)
     suspend fun logInWithApple(idToken: String)
-
-    fun logOut()
```
cristhianescobar, line 24: @martyu we should be using the UserSessionUseCase to check if the user is logged in or to log the user out.  I have added the `postNotification(Notification.loginStateChanged)` to the UserSessionUseCaseImpl

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/WalletNFTTracksUseCase.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.koin.core.component.inject
+
+
+/**
+ * `WalletNFTTracksUseCase` defines the contract for retrieving Non-Fungible Token (NFT) tracks associated with a wallet.
+ *
+ * This interface provides methods to retrieve a list of NFT tracks either as a continuous stream (Flow)
+ * or as a single suspending function call.
+ */
+interface WalletNFTTracksUseCase {
```
cristhianescobar, line 17: Rename from **WalletNFTSongsUseCase** to **WalletNFTTracksUseCase**

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/WalletNFTSongsUseCaseImpl.kt
```diff
@@ -2,19 +2,23 @@ package io.newm.shared.internal.implementations
...
 internal class WalletNFTSongsUseCaseImpl(
     private val cardanoRepository: CardanoWalletRepository,
     private val connectWalletManager: ConnectWalletManager
-) : WalletNFTSongsUseCase {
+) : WalletNFTTracksUseCase {
 
-    override fun getAllWalletNFTSongs(): Flow<List<NFTTrack>> = MockSongs.getSongsFlow()
-    override suspend fun getWalletNFTs(): List<NFTTrack> {
+    override fun getAllNFTTracksFlow(): Flow<List<NFTTrack>> {
+        return flow {
```
martyu, line 17: Can these flows be moved to the callsite in the view models since we can't use them on iOS, or does it need to be here for some reason?

cristhianescobar: We need them here, there will a web client down the road that can be power through KMM. Ideally once we figure out how to consume flows on iOS we can move to a more reactive flow.

android/app-newm/build.gradle.kts
```diff
@@ -86,20 +86,10 @@ android {
...
-    implementation(platform(Google.firebase))
-    implementation(project(Modules.coreTheme))
-    implementation(project(Modules.coreResources))
-    implementation(project(Modules.coreUiUtils))
-    implementation(project(Modules.login))
-    implementation(project(Modules.musicPlayer))
-    implementation(project(Modules.barcodeScanner))
-    implementation(project(Modules.shared))
-
+    implementation("com.jakewharton:process-phoenix:2.1.2")
```
wlara, line 89: Why not using the a dependency object string here like the other dependencies?

wlara (approved):

android/app-newm/build.gradle.kts
cristhianescobar: Need to do a dependency cleaning soon, will create a ticket to take care of that and a few others 👍🏼

PR #160, opened 2023-11-23
Makes player controller reusable

newmskywalker:


cristhianescobar (approved):
LFG! With this change do we know depoly to firebase directly?

newmskywalker:
> LFG! With this change do we know depoly to firebase directly?

We're about to find out!

PR #164, opened 2023-12-03
Skie

cristhianescobar:
Adding skie to enable easier interoperability between kotlin and swift, regarding seal classes


shared/src/commonMain/kotlin/io.newm.shared/internal/HttpRoutes.kt
```diff
@@ -1,5 +1,12 @@
...
 internal object HttpRoutes {
-    const val HOST = "https://garage.newm.io"
+    fun getHost(): String {
+        return if (isProduction) PROD_ENVIRONMENT else DEV_ENVIRONMENT
+    }
+
+    private const val DEV_ENVIRONMENT = "https://garage.newm.io"
+    private const val PROD_ENVIRONMENT = "https://studio.newm.io"
+
+    private const val isProduction: Boolean = false
```
martyu, line 11: Wondering if this should be a environment variable so we can switch it for CI and unit tests?

martyu (approved):

shared/src/commonMain/kotlin/io.newm.shared/internal/HttpRoutes.kt
cristhianescobar: Yes this will get moved into a file that can be configurable by CI

PR #165, opened 2023-12-04
Nft player

martyu:
https://projectnewm.atlassian.net/browse/MOBILE-74
https://projectnewm.atlassian.net/browse/MOBILE-72
https://projectnewm.atlassian.net/browse/MOBILE-75
https://projectnewm.atlassian.net/browse/MOBILE-60
https://projectnewm.atlassian.net/browse/MOBILE-61
https://projectnewm.atlassian.net/browse/MOBILE-62
https://projectnewm.atlassian.net/browse/MOBILE-73


iosApp/Modules/Helpers/AudioPlayer/PlayButton.swift
```diff
@@ -4,18 +4,29 @@ import Resolver
 import ModuleLinker
 
 public struct PlayButton: View {
-	@InjectedObject private var audioPlayer: AudioPlayerImpl
+	@InjectedObject private var audioPlayer: VLCAudioPlayer
```
cristhianescobar, line 7: What are your thoughts on using a protocol/interface for the AudioPlayer? This way, everywhere that requires an Audio player can use the protocol and have a separate VLCAudioPlayer that implements the AudioPlayer interface. I believe this would make it easier to scale and swap implementations as needed in the future. Additionally, it's probably better coding practice than having the code depend on the concrete implementation. I understand that for now, it may seem like unnecessary work, but it will be helpful down the road, thoughts?

iosApp/Modules/Helpers/Files/DownloadManager.swift
```diff
@@ -0,0 +1,78 @@
...
+		task.resume()
+	}
+	
+	func download(url: URL, progressHandler: @escaping (Double) -> Void) async throws -> URL {
+		try await withCheckedThrowingContinuation { [weak self] continuation in
+			guard let self else { fatalError() }
+			download(url: url) { result in
+				continuation.resume(with: result)
+			} progressHandler: { progress in
+//				print("\n\n\nCANCELLED\n\n\n")
```
cristhianescobar: Should we remove commented out code?

iosApp/Modules/Screens/Artist/Mocks/MockData.swift
```diff
@@ -14,13 +14,16 @@ class MockArtistViewUIModelProviding: ArtistViewUIModelProviding {
...
-			profileImage: URL(string: MockData.url(for: Asset.MockAssets.artist0))!,
+			profileImage: URL(string: "" /*MockData.url(for: Asset.MockAssets.artist0)*/)!,
 			followSection: SupportButton.followButton(),
 			supportSection: SupportButton.supportButton(),
-			trackSection: CellsSectionModel(cells: MockData.bigSongCells_shuffled(seed: 1, onTap: { id in
-				actionHandler.songTapped(id: id)
-			}), title: "LATEST TRACKS"),
-			topSongs: CellsSectionModel(cells: MockData.bigSongCells_shuffled(seed: 2, onTap: { id in actionHandler.songTapped(id: id) }), title: "TOP SONGS"),
+			trackSection: CellsSectionModel(cells: [],
+//											MockData.bigSongCells_shuffled(seed: 1, onTap: { id in
```
cristhianescobar: Same here?

iosApp/Modules/Screens/Home/HomeTests/HomeViewModelTests.swift
```diff
@@ -5,16 +5,16 @@ import AudioPlayer
...
-		XCTAssert(vm.route.isEmpty)
-		vm.artistTapped(id: "Joey Fidelity")
-		guard case .artist = vm.route.last else { return XCTFail() }
-		vm.route.removeLast()
-		vm.songTapped(id: "My Heart Hurts So Bad")
-		XCTAssertTrue(audioPlayer.playbackInfo.isPlaying)
-		XCTAssert(audioPlayer.song?.id == "My Heart Hurts So Bad")
-	}
+//	@InjectedObject var audioPlayer: VLCAudioPlayer
+//
```
cristhianescobar: Same

iosApp/Modules/Screens/Home/UI/Home Views/HomeViewModel.swift
```diff
@@ -12,7 +12,7 @@ class HomeViewModel: ObservableObject {
 	@Published var shouldShowGreeting: Bool = true
 	
 	private var uiModelProvider: HomeViewUIModelProviding { MockHomeViewUIModelProvider(actionHandler: self) }
-	@Injected private var audioPlayer: AudioPlayerImpl
+	@InjectedObject private var audioPlayer: VLCAudioPlayer
```
cristhianescobar, line 15: If we decide to go with the protocol could we just inject `AudioPlayer` instead here  and everywhere is needed?

iosApp/Modules/Screens/Library/LibraryModule.swift
```diff
@@ -1,22 +1,72 @@
...
+					fatalError()
+				}
 
+				func getWalletNFTs() async throws -> [NFTTrack] {
+					NFTTrack.mockTracks
+				}
+			}
+		}
+		
+//		mockResolver.register {
```
cristhianescobar: Can we remove this as well?

iosApp/Modules/Screens/Library/UI/LIbraryView.swift
```diff
@@ -2,31 +2,192 @@ import Foundation
...
+				}
+				.padding(.leading, 4)
+				
+				Spacer()
+				
+				progressView(for: track)
+			}
+			.foregroundStyle(.white)
+			.tag(track.id)
+		}
```
cristhianescobar, line 151: Same

iosApp/Modules/Screens/Login/UI/ViewModels/LandingViewModel.swift
```diff
@@ -79,7 +78,7 @@ class LandingViewModel: ObservableObject {
 //				try await userRepo.resetPassword(email: email, password: password, confirmPassword: confirmPassword, authCode: confirmationCode)
```
cristhianescobar: Is this removable also?

iosApp/Modules/Screens/Profile/ProfileMoreView.swift
```diff
@@ -0,0 +1,68 @@
+//import SwiftUI
+//import Fonts
+//import Colors
+//
+//struct ProfileMoreView: View {
```
cristhianescobar: Can we delete this as well?

iosApp/Modules/Screens/Profile/ProfileMoreViewModel.swift
```diff
@@ -0,0 +1,29 @@
+//import Foundation
+//import Resolver
+//import shared
+//
+//@MainActor
+//class ProfileMoreViewModel: ObservableObject {
```
cristhianescobar: Delete?

iosApp/Modules/Screens/Profile/ProfileView.swift
```diff
@@ -41,10 +45,35 @@ struct ProfileView: View {
 			}
 			.loadingToast(isLoading: $viewModel.isLoading)
 		}
-		.sheet(isPresented: $showMore) {
-			ProfileMoreView()
-				.presentationDetents([.medium])
+		//		.sheet(isPresented: $showMore) {
+		//			ProfileMoreView()
```
cristhianescobar: Same

iosApp/Modules/Screens/Profile/ProfileView.swift
```diff
@@ -54,42 +83,39 @@ struct ProfileView: View {
...
-			Button {
-				showMore = true
-			} label: {
-				Image(systemName: "ellipsis")
-					.rotationEffect(.degrees(90))
-					.foregroundColor(.white)
-			}
+//			Button {
+//				showMore = true
+//			} label: {
```
cristhianescobar: Same?

shared/src/commonMain/kotlin/io.newm.shared/internal/services/CardanoWalletAPI.kt
```diff
@@ -6,24 +6,29 @@ import io.ktor.client.request.get
...
-            parameter("xpub", xpub)
-        }.body()
+        try {
+            authClient.get("/v1/cardano/nfts") {
+                contentType(ContentType.Application.Json)
+                parameter("xpub", xpub)
+            }.body()
+        } catch(e: JsonConvertException) {
+            print("Error fetching wallet NFTs: ${e.message}")
+            emptyList()
```
cristhianescobar: I think we might want to revert this change, considering the possibility of errors occurring. In the case of errors, we should display an error state. Conversely, if a wallet genuinely has no NFTs, we should show an empty state.

shared/src/commonMain/kotlin/io.newm.shared/di/NetworkClientFactory.kt
```diff
@@ -110,12 +120,14 @@ internal class NetworkClientFactory(
...
+                                logOutUseCase.logout()
                                 logger.d { "NewmKMM - refreshTokens Invalid Token response: $renewTokens" }
-                                throw Exception("Invalid Token response")
+                                throw KMMException("Invalid Token response")
                             }
                         } catch (e: Exception) {
+                            logOutUseCase.logout()
                             logger.d { "NewmKMM - refreshTokens: Exception: $e" }
-                            null
+                            throw KMMException("Refresh token failed: $e")
```
cristhianescobar, line 130: Good catch here.

iosApp/Modules/Helpers/AudioPlayer/PlayButton.swift
martyu: Definitely don't like it like this, but there's an issue where you can't observe protocols in swift.  Causes a compiler error.  So I had split out the observation stuff into a separate read-only observable object, but it ended up being messier and still wouldn't allow for swapping implementations.  It's kind of a ridiculous language limitation.  I don't think it will be any trouble swapping them, just will need to do a find and replace.  Any it is still only relying on the public interface, since it's in a different module it doesn't have access to the internal stuff.  I could probably bundle all the observable stuff into a published struct on the audio player, and use that for observation.  But I'll look into that once it becomes necessary.

iosApp/Modules/Helpers/Files/DownloadManager.swift
martyu: I left this in on purpose because I still need to handle cancellation.  Although are you handling it on Android?  Do you ever cancel a download that's started?  I was going to make it so starting another download canceled the current one, but I ended up not.

iosApp/Modules/Screens/Artist/Mocks/MockData.swift
martyu: I'm leaving all these unused screen modules alone for when/if we bring them back.  I only commented out what was needed to silence compiler errors.  But I removed a lot of these from the build target so they're not being compiled anyway, so yeah I can uncomment it.

iosApp/Modules/Screens/Library/LibraryModule.swift
martyu: I can uncomment it, it's not being used anyway though since the actual implementation overrides it.

iosApp/Modules/Screens/Library/UI/LIbraryView.swift
martyu: Yeah I guess we don't need swipe anymore.

iosApp/Modules/Screens/Profile/ProfileMoreView.swift
martyu: I assumed this is coming back at some point?  Legal disclaimers live here.

shared/src/commonMain/kotlin/io.newm.shared/internal/services/CardanoWalletAPI.kt
martyu: I think this was crashing because of not handling expired refresh tokens...?  But yeah it should be reverted.

martyu: Actually I remember, it was crashing because we get back a 500 error from the server when the xPub is invalid.  How should we handle that case?

cristhianescobar: Still an error from, so we can display an error state for now, that 500 error was happening because we were missing the proper configuration

iosApp/Modules/Helpers/Files/DownloadManager.swift
cristhianescobar: I have not done the implementation for file download just yet, what would be the case where you cancel a download and start a new download? Are you talking about the user cancelling the download purposely?

iosApp/Modules/Screens/Profile/ProfileMoreView.swift
cristhianescobar: Maybe, maybe not. I usually end up deleting code that is not currently used

iosApp/Modules/Helpers/Files/DownloadManager.swift
martyu: That's another case yeah, but i was thinking if they started another download before the current one is finished, it would cancel the current, so there was no queue.  But now that I think of it, that seems silly.  So I'll just remove this.

iosApp/Modules/Screens/Profile/ProfileMoreView.swift
martyu: yeah, can always restore it from git if needed.  i'll remove.

shared/src/commonMain/kotlin/io.newm.shared/internal/services/CardanoWalletAPI.kt
martyu: Should we do a check when they enter their xPub key that it's valid, instead of allowing them to successfully add an invalid key?

cristhianescobar: We can do a check on the scanner, so that we only take the address with the xpub prefix

cristhianescobar (approved):

shared/src/commonMain/kotlin/io.newm.shared/internal/services/CardanoWalletAPI.kt
martyu: doesn't verify it's an actual wallet though.  we can come back to this.

PR #169, opened 2023-12-15
Newm app pointing to production

cristhianescobar:
- Adding production endpoint as default
- Replaced newm chain endpoint query with [new](https://github.com/projectNEWM/newm-server/wiki/GetCardanoWalletNFTSongs) endpoint
- Added the functionality to search songs
- Removed the testnet default wallet address


android/app-newm/src/main/java/io/newm/screens/library/screens/EmptyWalletScreen.kt
```diff
@@ -0,0 +1,54 @@
...
+        Column(horizontalAlignment = Alignment.CenterHorizontally) {
+            Text(
+                text = "No NFTs yet?",
+                style = MaterialTheme.typography.h6,
+                textAlign = TextAlign.Center,
+                modifier = Modifier.padding(16.dp)
+            )
+            Spacer(modifier = Modifier.height(8.dp))
+            Text(
+                text = "You can buy NFTs from our marketplace",
```
martyu, line 40: I'll need to add this to iOS too.  Added jira ticket.

android/app-newm/src/main/java/io/newm/Logout.kt
```diff
@@ -1,36 +1,48 @@
...
-            userSession.logout()
-            restartApp.run()
-            logger.d("Logout successful")
-            0
-        } catch (e: Exception) {
-            logger.d("Logout Exception: $e")
-            1
-        }
+    fun signOutUser() = try {
+        loginUseCase.logout()
```
martyu, line 24: probably should clean the song file cache when logging out too?

martyu (approved):
If the songs are playing on android then we can merge this for now and i'll check ios once I get my wallet filled.

android/app-newm/src/main/java/io/newm/screens/library/screens/EmptyWalletScreen.kt
cristhianescobar: This is a placeholder, waiting to hear back from design about what the screen should look like and say.

android/app-newm/src/main/java/io/newm/Logout.kt
cristhianescobar: The logout flows gets started on the client, so that is work we need to do before calling the `loginUseCase.logout()`

Android has not implemented the caching mechanism yet.

martyu: But you download the song file don't you?  You're not streaming it?

PR #174, opened 2024-01-06
MOBILE-96 | Add swipe to download button to Library Screen

Jermainelr:
download_song


newmskywalker (approved):

cristhianescobar:
I will go ahead and move this change into the NFT library tab, but going forward let's make sure those changes also reflect into the NFT library as well

PR #175, opened 2024-01-11
Cache current user

cristhianescobar:
- Adding a table for current user so we don't have to fetch from network every time the account page is loaded.
- Also created a separate section for NFT Tracks vs Stream Tokens
Screenshot 2024-01-10 at 11 40 37 PM


Jermainelr (approved):
LGTM

PR #177, opened 2024-01-20
Add non-Flow kmm functions for iOS.

martyu:
Add non-Flow kmm functions for iOS.
fixed error alerts


shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/WalletNFTSongsUseCaseImpl.kt
```diff
@@ -13,12 +14,19 @@ internal class WalletNFTSongsUseCaseImpl(
         return cardanoRepository.getWalletCollectableTracks()
     }
 
-    override fun getAllStreamTokens(): Flow<List<NFTTrack>> {
```
cristhianescobar, line 16: Can we keep the function and just rename it so we can have a flow and a non flow version?
`override fun getAllStreamTokensFlow(): Flow<List<NFTTrack>>`
`override fun getAllStreamTokens(): List<NFTTrack>`

Also, at what point did flows stop working? I was under the impression that you were able to consume them without a problem?

cristhianescobar (approved):

PR #178, opened 2024-01-21
Account Screen update

cristhianescobar:
Screenshot 2024-01-21 at 1 29 06 AM
Screenshot 2024-01-21 at 1 29 27 AM


martyu:
Should iOS look like this too?

android/core/resources/src/main/res/values/strings.xml
```diff
@@ -1,75 +1,81 @@
...
-    <string name="tribe">Tribe</string>
-    <string name="stars">Stars</string>
-    <string name="wallet">Wallet</string>
+    <string name="library_albums">YOUR ALBUMS</string>
+    <string name="library_artists">YOUR ARTISTS</string>
+    <string name="library_download">Download</string>
+    <string name="library_empty_content_subtitle">Time to rescue it with your epic music stash! 🎶</string>
+    <string name="library_empty_content_title">Your library is empty.</string>
+    <string name="library_search">Search</string>
+    <string name="logout_dialog_message">Are you sure you want to logout?</string>
```
martyu:
```suggestion
    <string name="logout_dialog_message">Are you sure you want to log out?</string>
```

android/core/resources/src/main/res/values/strings.xml
```diff
@@ -1,75 +1,81 @@
...
+    <string name="logout_dialog_title">👋 %s not so fast!</string>
```
martyu:
```suggestion
    <string name="logout_dialog_title">👋 %s, not so fast!</string>
```

martyu: why are we waving goodbye as we're trying to stop them from leaving...? 😸

shared/build.gradle.kts
```diff
@@ -33,7 +33,7 @@ kotlin {
 
     val xcf = XCFramework()
     listOf(
-//        iosX64(),
+        iosX64(),
```
martyu, line 36: Did you mean to uncomment this?

martyu (dismissed):

martyu:
two little grammar fixes then looks good

android/core/resources/src/main/res/values/strings.xml
cristhianescobar: I just put something together for the internal team to see, the copy might change.

shared/build.gradle.kts
cristhianescobar: Yeah, was cleaning up the share gradle file and since we need this long term I put it back.

cristhianescobar:
> Should iOS look like this too?

It would be nice, I basically just put something together for the team to see.

newmskywalker (approved):

PR #180, opened 2024-01-22
Make songs play in-line instead of opening a new screen

newmskywalker:


shared/src/commonMain/kotlin/io.newm.shared/internal/HttpRoutes.kt
```diff
@@ -8,5 +8,5 @@ internal object HttpRoutes {
     private const val DEV_ENVIRONMENT = "https://garage.newm.io"
     private const val PROD_ENVIRONMENT = "https://studio.newm.io"
 
-    private const val isProduction: Boolean = true
-}
\ No newline at end of file
+    private const val isProduction: Boolean = false
```
cristhianescobar: Can we revert this one to 'true'? That way, when we release the build to the team, they can test the connection on the main chain. Since we're doing read-only operations, we should be fine.

cristhianescobar (approved):
Looks good to me, just update the boolean to prod for now. We probably need to add a way to pick between mainet and testnet down the road

PR #185, opened 2024-01-31
Refreshing and error handling

martyu:
Move Utilities into its own module.
Add SharedExtensions module for bridging the shared kmm module with iOS.
Add tests.
Add refresh for fetching nft tracks.
Show error message from api.


shared/build.gradle.kts
```diff
@@ -1,112 +1,134 @@
...
-        iosSimulatorArm64()
-    ).forEach {
-        it.binaries.framework {
-            baseName = "shared"
-            xcf.add(this)
-        }
-    }
+	val xcf = XCFramework()
+	listOf(
+//        iosX64(),
```
cristhianescobar, line 36: Can we remove this, since is part of the old architecture?

shared/build.gradle.kts
```diff
@@ -1,112 +1,134 @@
...
+				implementation(Ktor.iosDarwin)
+				implementation(SqlDelight.nativeDriver)
+			}
+		}
+//        val iosX64Test by getting
+//        val iosArm64Test by getting
+		val iosSimulatorArm64Test by getting
+		val iosTest by creating {
+			dependsOn(commonTest)
+//            iosX64Test.dependsOn(this)
```
cristhianescobar, line 106: Same here

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/LoginUseCaseImpl.kt
```diff
@@ -1,41 +1,54 @@
...
     override suspend fun logInWithApple(idToken: String) {
-        return repository.oAuthLogin(OAuthData.Apple(idToken))
+        return mapErrorsSuspend {
+            return@mapErrorsSuspend repository.oAuthLogin(OAuthData.Apple(idToken))
+        }
     }
 
     override fun logout() {
+        connectWalletUseCase.disconnect()
         repository.logout()
```
cristhianescobar, line 52: The repository.logout() takes care of disconnecting the wallet
```
 fun logout() {
        tokenManager.clearToken()
        connectWalletManager.disconnect()
    }
```

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/utilities/MapErrors.kt
```diff
@@ -0,0 +1,28 @@
...
+    try {
+        return block()
+    } catch (e: KMMException) {
+        throw e
+    } catch (e: CancellationException) {
+        throw e
+    } catch (e: Throwable) {
+        throw KMMException("We encountered a problem", e)
+    }
+}
```
cristhianescobar, line 28: What do you think about this @newmskywalker, can you think of a cleaner alternative?

```
suspend fun <T> runCatchingAsyc(block: suspend () -> T): T {
    return runCatching { block() }
        .getOrElse { e ->
            when (e) {
                is KMMException, is CancellationException -> throw e
                else -> throw KMMException("We encountered a problem: ${e.message}", e)
            }
        }
}
```
```
fun <T> runCatchingSync(block: () -> T): T {
    return runCatching { block() }
        .getOrElse { e ->
            when (e) {
                is KMMException, is CancellationException -> throw e
                else -> throw KMMException("We encountered a problem: ${e.message}", e)
            }
        }
}
```

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/LoginUseCaseImpl.kt
martyu: I moved it here.  We want the disconnect use case called so any side effects can run.

cristhianescobar: I thought we had agreed on opting against the practice of calling use cases within use cases. This decision was made with the desire to uphold a clear separation of concerns, so that each use case is distinctly responsible for its unique business function. This not only enhances the readability and maintainability of our code, but also greatly simplifies the testing process. By enabling each use case to be tested independently, without the intricacies of intertwined dependencies, we're promoting a more modular and robust system.

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/utilities/MapErrors.kt
cristhianescobar: Since the function itself can be called from a coroutine asynchronously, we don't need the runCatchingAsync version of this.

shared/src/commonMain/kotlin/io.newm.shared/internal/services/models/Users.kt
```diff
@@ -0,0 +1,62 @@
...
+internal data class GoogleSignInRequest(val accessToken: String)
+
+@Serializable
+internal data class AppleSignInRequest(val idToken: String)
+
+@Serializable
+internal data class FacebookSignInRequest(val accessToken: String)
+
+@Serializable
+internal data class LinkedInSignInRequest(val accessToken: String)
```
cristhianescobar, line 21: Can we make this instead
```
@Serializable
sealed class SignInRequest

@Serializable
internal data class GoogleSignInRequest(val accessToken: String) : SignInRequest()

@Serializable
internal data class AppleSignInRequest(val idToken: String) : SignInRequest()

@Serializable
internal data class FacebookSignInRequest(val accessToken: String) : SignInRequest()

@Serializable
internal data class LinkedInSignInRequest(val accessToken: String) : SignInRequest()
```

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/WalletNFTTracksUseCase.kt
```diff
@@ -24,6 +26,8 @@ interface WalletNFTTracksUseCase {
      * @return Flow<List<NFTTrack>> - A flow emitting the list of NFT tracks associated with the wallet.
      */
     fun getAllNFTTracksFlow(): Flow<List<NFTTrack>>
+
+    @Throws(KMMException::class, CancellationException::class)
```
cristhianescobar, line 30: Why the we remove this annotations from the API level?

iosApp/Modules/Mocks/Use Cases/MockChangePasswordUseCase.swift
```diff
@@ -0,0 +1,23 @@
...
+	public var errorToThrow: Error?
+	
+	public init(errorToThrow: Error? = nil) {
+		self.errorToThrow = errorToThrow
+	}
+	
+	public func changePassword(oldPassword: String, newPassword: String, confirmPassword: String) async throws {
+		if let errorToThrow {
+			throw errorToThrow
+		}
```
cristhianescobar, line 16:
```suggestion hi
		if let errorToThrow {
			throw errorToThrow
		}
```

cristhianescobar (approved):

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/LoginUseCaseImpl.kt
martyu: discussed on a call, leaving it here so can call ConnectWalletUseCase and not miss any side effects.

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/WalletNFTTracksUseCase.kt
martyu: only needed at ios-kmm boundary

PR #208, opened 2024-02-20
Mobile Recaptcha

cristhianescobar:
Android and iOS Recaptcha Implementation


shared/src/commonMain/kotlin/io.newm.shared/config/NewmSharedBuildConfig.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    val walletConnectProjectId: String
+        get() = SharedBuildConfig.WALLET_CONNECT_PROJECT_ID
+
+    val recaptchaSiteKey: String
+        get() = SharedBuildConfig.RECAPTCHA_SITE_KEY
+
+    val isDevMode: Boolean
+        get() = mode == Mode.DEVELOPMENT
+}
```
martyu: Why are these in kmm if they're not used in kmm?  (besides walletConnectProjectId)

martyu: iOS needs some of these at compile time, but this object is only available at runtime.

cristhianescobar: I am exposing them through KMP, so both clients can have access to it, these values are available at compile time, there is code that gets generated before the app actually runs.

android/app-newm/src/main/java/io/newm/LoginActivity.kt
```diff
@@ -112,6 +121,20 @@ class LoginActivity : ComponentActivity() {
...
+        lifecycleScope.launch {
+            Recaptcha.getClient(application, NewmSharedBuildConfig.recaptchaSiteKey, timeout = 20000L)
+                .onSuccess { client ->
+                    recaptchaClientProvider.setRecaptchaClient(client)
+                }
+                .onFailure { exception ->
+                    // Handle communication errors ...
+                    // See "Handle communication errors" section
+                    Log.e("LoginActivity", "Setup failed", exception)
+                }
```
newmskywalker, line 135: should we just crash? Or can the app function without this?

android/features/login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountScreenPresenter.kt
```diff
@@ -65,9 +69,14 @@ class CreateAccountScreenPresenter constructor(
...
-                                    signupUseCase.requestEmailConfirmationCode(
-                                        email = userEmail.text,
-                                    )
+                                    recaptchaClientProvider.get().execute(RecaptchaAction.SIGNUP).onSuccess { token ->
+                                        signupUseCase.requestEmailConfirmationCode(
+                                            email = userEmail.text,
+                                            humanVerificationCode = token
+                                        )
+                                    }.onFailure {
+                                        errorMessage = "Are you even a human?"
```
newmskywalker, line 78: should we reset the state to `EmailAndPassword` on falure?

android/features/login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountScreenPresenter.kt
```diff
@@ -94,15 +103,19 @@ class CreateAccountScreenPresenter constructor(
...
+                                            email = userEmail.text,
+                                            verificationCode = verificationCode.text,
+                                            password = password.text,
+                                            passwordConfirmation = passwordConfirmation.text,
+                                            nickname = name.text,
+                                            humanVerificationCode = token,
+                                        )
+                                        navigateHome()
+                                    }.onFailure {
+                                      errorMessage = "Are you even a human?"
```
newmskywalker, line 117: same here

android/features/login/src/main/java/io/newm/feature/login/screen/resetpassword/ResetPasswordScreenPresenter.kt
```diff
@@ -56,8 +60,12 @@ class ResetPasswordScreenPresenter(
                             isLoading = true
                             coroutineScope.launch {
                                 try {
-                                    signupUseCase.requestEmailConfirmationCode(email.text)
-                                    step = ResetPasswordStep.EnterVerificationCode
+                                    recaptchaClientProvider.get().execute(RecaptchaAction.custom("RequestCode")).onSuccess { token ->
+                                        signupUseCase.requestEmailConfirmationCode(email.text, humanVerificationCode = "")
```
newmskywalker: is it intentional that we're not using the token here?

android/features/login/src/main/java/io/newm/feature/login/screen/resetpassword/ResetPasswordScreenPresenter.kt
```diff
@@ -104,6 +112,7 @@ class ResetPasswordScreenPresenter(
                                             code = authCode.text,
                                             newPassword = password.text,
                                             confirmPassword = passwordConfirmation.text,
+                                            humanVerificationCode = ""
```
newmskywalker: do we need to wrap this with the captcha service? or is that not needed?

newmskywalker (approved):

android/app-newm/src/main/java/io/newm/LoginActivity.kt
cristhianescobar: We need to add Sentry to report crashes. That's my next task. However, we might consider doing that, although I think we should restart the app if the user encounters a failure here.

android/features/login/src/main/java/io/newm/feature/login/screen/resetpassword/ResetPasswordScreenPresenter.kt
cristhianescobar: Good catch, this needs to send the token as well

android/features/login/src/main/java/io/newm/feature/login/screen/resetpassword/ResetPasswordScreenPresenter.kt
cristhianescobar: Same thanks for the catch.

shared/src/commonMain/kotlin/io.newm.shared/config/NewmSharedBuildConfig.kt
martyu: How did you verify they're available at compile time in iOS as an environment variable?  That's how they're used.

cristhianescobar: I ran the iOS app on Android studio and it worked fine. You don't have to use the same variables if you don't like, you can continue using the secrets file for now.

PR #219, opened 2024-04-02
MOBILE-179

martyu:
Pull out "play queue" functionality from VLCAudioPlayer

Add "offline dev mode" with stubs (the plane didn't have internet so i threw this in on the flight to unblock.)


iosApp/Modules/Screens/Library/UI/FilterView.swift
```diff
@@ -0,0 +1,38 @@
+//import SwiftUI
+//import shared
+//import AudioPlayer
+//
+//struct FilterSortMenuView: View {
```
cristhianescobar: Can we remove the code we are using here?

iosApp/Modules/Screens/NowPlaying/UI/NowPlayingView.swift
```diff
@@ -24,6 +24,14 @@ public struct NowPlayingView: View {
 		}
 		.background(background)
 		.backButton()
+		//TODO: make an "ErrorProviding" protocol, replace all these.
```
cristhianescobar, line 27: same here, can this be remved?

iosApp/Modules/Mocks/MockData/Music/getSchwifty1.mp3
cristhianescobar, line 1: Can we remove these tracks, or at least ensure they are not being deployed with the app when we send out a build. We can only play tracks owned by the user's wallet.

cristhianescobar (changes requested):
Can we ensure the hard-coded tracks are not available to user? We should only play tracks they owned. If this is just for testing that is fine, we just cannot ship this tracks out as we don't as we cannot pay royalties from the app yet.

iosApp/Modules/Screens/Library/UI/FilterView.swift
martyu: bah, yep, that was the chatgpt experiment.  it did ok 😐

iosApp/Modules/Screens/NowPlaying/UI/NowPlayingView.swift
martyu: I still need to add this back in but there were issues with the audio player errors I need to fix first.

iosApp/Modules/Mocks/MockData/Music/getSchwifty1.mp3
martyu: They're only in debug builds

martyu: Only there so we can continue development when there's no internet (I use it on the plane sometimes)

PR #223, opened 2024-05-05
Feature Branch - Newm Wallet Connection Flow Integration

cristhianescobar:
Implementation of the newm wallet connection flow
-  I have set the environment to staging for now, as the new wallet flow is not in production yet

@martyu when you get a chance can you make the necessary changes on iOS for the new flow to work, we need those changes prior to merging this PR.


shared/src/commonMain/kotlin/io.newm.shared/public/usecases/ConnectWalletUseCase.kt
```diff
@@ -1,49 +1,51 @@
...
 import org.koin.core.component.inject
 
 /**
  * `ConnectWalletUseCase` defines the contract for managing wallet connections.
  *
- * This interface provides methods to connect to a wallet using an extended public key (xpub),
+ * This interface provides methods to connect to a wallet using a wallet connection id,
  * disconnect from the wallet, and check if a wallet is currently connected.
  */
 interface ConnectWalletUseCase {
```
cristhianescobar, line 13: @martyu the changes on this class is what IOS should be concern with.

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ConnectWalletUseCaseImpl.kt
```diff
@@ -1,37 +1,40 @@
...
 internal class ConnectWalletUseCaseImpl(
-    private val connectWalletManager: ConnectWalletManager,
     private val cardanoWalletRepository: CardanoWalletRepository
 ) : ConnectWalletUseCase {
-    override fun connect(xpub: String) {
+    override fun connect(walletConnectionId: String) {
         mapErrors {
-            connectWalletManager.connect(xpub)
             postNotification(Notification.walletConnectionStateChanged)
+            cardanoWalletRepository.connectWallet(walletConnectionId)
```
martyu: Should change it before sending the notification I think?
```suggestion
            cardanoWalletRepository.connectWallet(walletConnectionId)
            postNotification(Notification.walletConnectionStateChanged)
```

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ConnectWalletUseCaseImpl.kt
```diff
@@ -1,37 +1,40 @@
...
         }
     }
 
-    override fun disconnect() {
+    override fun disconnect(walletConnectionId: String?) {
         mapErrors {
-            connectWalletManager.disconnect()
             cardanoWalletRepository.deleteAllNFTs()
             postNotification(Notification.walletConnectionStateChanged)
+            cardanoWalletRepository.disconnectWallet(walletConnectionId)
```
martyu: Same here
```suggestion
            cardanoWalletRepository.disconnectWallet(walletConnectionId)
            postNotification(Notification.walletConnectionStateChanged)
```

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ConnectWalletUseCaseImpl.kt
```diff
@@ -1,37 +1,40 @@
...
+import kotlinx.coroutines.flow.map
 import shared.Notification
 import shared.postNotification
 
 internal class ConnectWalletUseCaseImpl(
-    private val connectWalletManager: ConnectWalletManager,
     private val cardanoWalletRepository: CardanoWalletRepository
 ) : ConnectWalletUseCase {
-    override fun connect(xpub: String) {
+    override fun connect(walletConnectionId: String) {
```
martyu: Won't this be `suspend` now since it makes an api call?

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
```diff
@@ -100,15 +119,25 @@ internal class CardanoWalletRepository(
...
         } catch (e: Exception) {
             logger.e(e) { "Error fetching NFTs from network ${e.cause}" }
             throw e
         }
     }
 
+    private suspend fun fetchWalletConnections() {
+        try {
+            val walletConnections = walletConnectionAPI.getWalletConnections()
+            logger.d { "cje466 Fetched wallet connections from network: $walletConnections" }
```
martyu: cje466 username here

shared/src/commonMain/kotlin/io.newm.shared/config/NewmSharedBuildConfig.kt
```diff
@@ -29,7 +29,7 @@ object NewmSharedBuildConfig: KoinComponent {
     init {
         // If there's no mode saved in KVault, default to Production.
         if (storage.string(APP_MODE) == null) {
-            mode = Mode.PRODUCTION
+            mode = Mode.STAGING
```
martyu: did you mean to commit this?

android/app-newm/src/main/java/io/newm/screens/library/screens/LinkWalletScreen.kt
```diff
@@ -35,9 +35,9 @@ fun LinkWalletScreen(onConnectWallet: (String) -> Unit) {
             // Handle the returned result here
             val data = result.data
             // Do something with the data
-            val xpubKey = data?.getStringExtra(BarcodeScannerActivity.XPUB_KEY).orEmpty()
-            Toast.makeText(context, "Wallet connected $xpubKey", Toast.LENGTH_SHORT).show()
-            onConnectWallet(xpubKey)
+            val newmWalletConnectionId = data?.getStringExtra(BarcodeScannerActivity.NEWM_KEY).orEmpty()
```
martyu: Might be good to use newmWalletConnectionId/NEWM_WALLET_CONNECTION_ID everywhere instead of NEWM_KEY

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ConnectWalletUseCaseImpl.kt
```diff
@@ -1,37 +1,40 @@
...
         mapErrors {
-            connectWalletManager.connect(xpub)
             postNotification(Notification.walletConnectionStateChanged)
+            cardanoWalletRepository.connectWallet(walletConnectionId)
         }
     }
 
-    override fun disconnect() {
+    override fun disconnect(walletConnectionId: String?) {
```
martyu: Same, `suspend` I think?

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/ConnectWalletUseCase.kt
```diff
@@ -1,49 +1,51 @@
...
-     *
-     * This method is useful for observing changes in the connection status of the wallet in a reactive manner.
-     * It ensures that any updates to the connection status are reflected in real-time.
+     * Retrieves a flow indicating whether any wallet connections exist.
      *
-     * @return Flow<Boolean> - A flow emitting the current connection status of the wallet.
+     * @return A flow emitting `true` if there are wallet connections, `false` otherwise.
      */
-    fun isConnectedFlow(): Flow<Boolean>
+    @NativeCoroutines
```
martyu: These caused issues on iOS remember, we'll need non-flow versions of the flows.

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryViewModel.kt
```diff
@@ -29,8 +29,8 @@ class NFTLibraryViewModel(
     private val queryFlow = MutableStateFlow("")
 
     val state: StateFlow<NFTLibraryState> by lazy {
-        connectWalletUseCase.isConnectedFlow().flatMapLatest { isConnected ->
-            if (isConnected) {
+        connectWalletUseCase.getWalletConnections().flatMapLatest { walletConnectionsList ->
+            if (walletConnectionsList.isNotEmpty()) {
```
martyu: Could this just use `hasWalletConnections`?

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
```diff
@@ -128,4 +157,38 @@ internal class CardanoWalletRepository(
...
+                    id = connection.id,
+                    createdAt = connection.createdAt,
+                    stakeAddress = connection.stakeAddress
+                )
+            }
+        }
+    }
+
+    fun connectWallet(newmCode: String) {
+        scope.launch {
```
martyu: Why are we doing these `scope.launch`s instead of suspend?

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
```diff
@@ -100,15 +119,25 @@ internal class CardanoWalletRepository(
...
         }
     }
 
+    private suspend fun fetchWalletConnections() {
+        try {
+            val walletConnections = walletConnectionAPI.getWalletConnections()
+            logger.d { "cje466 Fetched wallet connections from network: $walletConnections" }
+            cacheWalletConnections(walletConnections)
+        } catch (e: Exception) {
+            logger.e(e) { " cje466 Error fetching wallet connections from network ${e.cause}" }
```
martyu: and here

martyu (changes requested):
Don't forget anything in KMM exposed to iOS needs the     @Throws(KMMException::class, CancellationException::class)
 annotation

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
cristhianescobar: Made the function suspended instead.

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryViewModel.kt
cristhianescobar: This file got removed

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/ConnectWalletUseCase.kt
cristhianescobar: That is what Android Studio wants me to annotate with :(

shared/src/commonMain/kotlin/io.newm.shared/config/NewmSharedBuildConfig.kt
cristhianescobar: The new wallet flow is only available in staging for now, that is why I left it there. But ideally each client should have a dev settings menu where we can toggle each env

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ConnectWalletUseCaseImpl.kt
cristhianescobar: Made the function suspended instead.

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/ConnectWalletUseCase.kt
martyu: What does that mean? We should just take out that KMP NativeCoroutines if it's causing warnings, we're not using it anyway.

shared/src/commonMain/kotlin/io.newm.shared/config/NewmSharedBuildConfig.kt
martyu: yeah, we don't want to just switch it on testers right?  it would log them out and they wouldn't know why.

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
martyu: Doesn't look like it...?

shared/src/commonMain/kotlin/io.newm.shared/config/NewmSharedBuildConfig.kt
cristhianescobar: Fair, We probably need to hold this PR until we have deployed the newm tools to production, that way it doesn't affect any users.

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/LoginUseCaseImpl.kt
```diff
@@ -47,7 +47,7 @@ internal class LoginUseCaseImpl(
         }
     }
 
-    override fun logout() {
+    override suspend fun logout() {
         connectWalletUseCase.disconnect()
```
martyu: We won't want to disconnect the wallet anymore right?

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ConnectWalletUseCaseImpl.kt
```diff
@@ -1,37 +1,51 @@
...
 import shared.Notification
 import shared.postNotification
 
 internal class ConnectWalletUseCaseImpl(
-    private val connectWalletManager: ConnectWalletManager,
     private val cardanoWalletRepository: CardanoWalletRepository
 ) : ConnectWalletUseCase {
-    override fun connect(xpub: String) {
+    override suspend fun connect(walletConnectionId: String) {
         mapErrors {
```
martyu:
```suggestion
        mapErrorsSuspend {
```

martyu:
@cristhianescobar did you mean to add the kvault stuff here?  And force push...?

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
martyu: is this going to be suspend?

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ConnectWalletUseCaseImpl.kt
```diff
@@ -1,37 +1,57 @@
...
-            connectWalletManager.connect(xpub)
+            cardanoWalletRepository.connectWallet(walletConnectionId)
             postNotification(Notification.walletConnectionStateChanged)
+            cardanoWalletRepository.connectWallet(walletConnectionId)
```
martyu:
```suggestion
            cardanoWalletRepository.connectWallet(walletConnectionId)
            postNotification(Notification.walletConnectionStateChanged)
            cardanoWalletRepository.connectWallet(walletConnectionId)
```
```suggestion
            cardanoWalletRepository.connectWallet(walletConnectionId)
            postNotification(Notification.walletConnectionStateChanged)
```

martyu (approved):
almost there...

martyu (dismissed):
sorry meant to request changes

shared/src/commonMain/kotlin/io.newm.shared/config/NewmSharedBuildConfig.kt
cristhianescobar: I have now changed this to production by default.

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
cristhianescobar: Yes this is suspended

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/CardanoWalletRepository.kt
cristhianescobar: removed

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/LoginUseCaseImpl.kt
cristhianescobar: Not anymore, since the wallet is connected we will keep it that way so next time they login there is a wallet associated with the account.

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryPresenter.kt
```diff
@@ -15,27 +16,38 @@ import io.newm.feature.musicplayer.rememberMediaPlayer
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 
 class NFTLibraryPresenter(
     private val navigator: Navigator,
+    private val hasWalletConnectionsUseCase: HasWalletConnectionsUseCase,
     private val connectWalletUseCase: ConnectWalletUseCase,
-    private val walletNFTTracksUseCase: WalletNFTTracksUseCase
+    private val walletNFTTracksUseCase: WalletNFTTracksUseCase,
+    private val scope: CoroutineScope,
```
newmskywalker, line 30: Are we injecting the scope so the connect outlives the presenter? 
Or would it make sense for the coroutine to get canceled when this presenter is discarded?

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/SyncWalletConnectionsUseCase.kt
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Synchronizes wallet connections from the network to the device's database.
+     * This method should be called at least once per app launch or after connecting a wallet
+     * to ensure the database is seeded from the network.
+     *
+     * @return A list of [WalletConnection] fetched from the network.
+     * @throws KMMException If an application-specific error occurs.
+     * @throws CancellationException If the operation is cancelled.
+     */
+    @Throws(KMMException::class, CancellationException::class)
```
newmskywalker, line 23: not for this pr, but I'm wondering if the use cases should handle errors instead of throwing exceptions.  I think it would reduce code duplication and we would have a single source of truth for error handling

newmskywalker (approved):

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryPresenter.kt
cristhianescobar: The idea is that we want the call to only be made once per app launch. Since this is a top level tab, I don't think this will ever get discarded, given there is only a handful of screens.

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/SyncWalletConnectionsUseCase.kt
cristhianescobar: iOS uses the exceptions thrown to understand what to do next. The paradigm that I see being used is to not return anything unless there is an error. For example, on the wallet disconnect, I would like to map the responses to a data model, specifically a sealed class, to switch on the different responses. However, I believe it is not very friendly to consume sealed classes on iOS. @martyu

PR #227, opened 2024-05-06
Newm wallet connection flow - ios

martyu:
I also removed `connectWalletUseCase.disconnect()` from `LoginUseCaseImpl.logout()` since it's connected to their account now, not their device.


android/app-newm/src/main/java/io/newm/screens/account/UserAccountViewModel.kt
```diff
@@ -0,0 +1,60 @@
...
+import io.newm.shared.public.usecases.UserDetailsUseCase
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+class UserAccountViewModel(
```
cristhianescobar: This class has been deleted. Can we remove it from here as well.

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryViewModel.kt
```diff
@@ -0,0 +1,128 @@
...
+import io.newm.shared.public.usecases.WalletNFTTracksUseCase
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+@FlowPreview
+class NFTLibraryViewModel(
```
cristhianescobar: Same here, this class has been removed.

cristhianescobar (changes requested):

cristhianescobar (approved):
LGTM

PR #236, opened 2024-05-15
Force App Update API Integration

cristhianescobar:
Integration of the remote config API to enable force app update when necessary. For both Android and iOS clients


shared/src/commonMain/kotlin/io.newm.shared/internal/services/RemoteConfigAPI.kt
```diff
@@ -0,0 +1,20 @@
...
+import io.newm.shared.internal.services.models.MobileConfig
+import org.koin.core.component.KoinComponent
+
+internal class RemoteConfigAPI(networkClient: NetworkClientFactory) : KoinComponent {
+
+    private val httpClient: HttpClient = networkClient.httpClient()
+
+    suspend fun getMobileConfig(): MobileConfig =
+        httpClient.get("v1/mobile-config") {
+            contentType(ContentType.Application.Json)
```
wlara: This should be `accept(...)` instead of `contentType(...)` since its an HTTP GET.

wlara (approved):

shared/src/commonMain/kotlin/io.newm.shared/internal/services/RemoteConfigAPI.kt
cristhianescobar: Good catch! Thanks

PR #241, opened 2024-05-23
Jetpack Preferences Datastore in Kotlin Multiplatform Mobile (KMM)

cristhianescobar:
Replacing third party library KVault with a Google supported library.

With this change we will be able to surface the Preferences Datastore to both android and ios clients through  Koin


shared/src/iosMain/kotlin/io.newm.shared/internal/implementations/PreferencesDataStoreImpl.ios.kt.kt
```diff
@@ -0,0 +1,39 @@
+package io.newm.shared.internal.implementations
+
+import io.newm.shared.internal.db.PreferencesDataStore
+import platform.Foundation.NSUserDefaults
+
+class PreferencesDataStoreImpl : PreferencesDataStore {
```
cristhianescobar, line 6: @marty can you double check this works well on IOS? I don't have a physical device to install the iOS app in

wlara (approved):

PR #244, opened 2024-05-30
Clean up profile links and open them in web browser

newmskywalker:


android/app-newm/src/main/java/io/newm/screens/profile/ProfileBottomSheetLayout.kt
```diff
@@ -57,41 +57,41 @@ fun ProfileBottomSheetLayout(
...
-                    Spacer(modifier = Modifier.height(8.dp))
-                    SecondaryButton(
-                        text = stringResource(R.string.profile_documents),
-                        textColor = White,
-                        onClick = onShowDocuments,
-
-                        )
-                    Spacer(modifier = Modifier.height(8.dp))
+//                    Text(
+//                        text = stringResource(id = R.string.profile_help_and_support),
```
cristhianescobar: Should we removed all this code for now since it's not being used anymore?

cristhianescobar (approved):
LGTM, minor comment on removing commented out code

PR #249, opened 2024-06-06
Add functionality to update profile and change password

newmskywalker:


shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/UserRepository.kt
```diff
@@ -39,30 +39,33 @@ internal class UserRepository(
...
-                        biography = user.biography,
-                        walletAddress = user.walletAddress,
-                        email = user.email,
-                        companyName = user.companyName,
-                        companyLogoUrl = user.companyLogoUrl,
-                        companyIpRights = user.companyIpRights,
-                        verificationStatus = user.verificationStatus,
-                    )
+                    dbWrapper().transaction {
+                        dbWrapper().userQueries.deleteAll() // just in case...
```
cristhianescobar, line 43: 👍🏼

android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditUi.kt
```diff
@@ -148,11 +161,25 @@ private fun ProfileScreenPreview() {
...
+                newPasswordState = TextFieldState(),
+                confirmPasswordState = TextFieldState(),
+                errorMessage = null,
                 eventSink = {},
             )
         )
     }
 }
+
+private fun User.toProfile(): Content.Profile {
```
cristhianescobar, line 178: Does it make sense to have both user and profile models?

android/app-newm/src/main/java/io/newm/screens/profile/ProfileForm.kt
```diff
@@ -68,26 +76,26 @@ fun ProfileForm(
                     style = formTitleStyle
                 )
                 Spacer(modifier = Modifier.height(16.dp))
-                TextFieldWithLabel(
-                    labelResId = R.string.profile_form_password_current_password,
-                    onValueChange = { value -> onProfileUpdated(profile.copy(currentPassword = value)) },
-                    isPassword = true,
-                    textfieldBackgroundColor = Gray23,
-                    placeholderResId = R.string.profile_placeholder_your_password
+                Password(
```
cristhianescobar, line 79: Do we need an email code to be able to update password? I assume we don't just double-checking

cristhianescobar (approved):
LGTM

android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditUi.kt
newmskywalker: I think it provides better separation of concerns. 
If we need to update the User class for whatever reason (or even delete it), we won't have to refactor the UI as well

android/app-newm/src/main/java/io/newm/screens/profile/ProfileForm.kt
newmskywalker: We don't, it's only needed to update the email

PR #250, opened 2024-06-07
Newm Connect Wallet On Android Up and Running

cristhianescobar:
- Refactor wallet connect code to have a use case per action
- Fully integrated new wallet flow
- Fix wallet removal


shared/src/androidMain/kotlin/io.newm.shared/internal/implementations/PreferencesDataStoreImpl.android.kt
```diff
@@ -64,4 +64,12 @@ class PreferencesDataStoreImpl(private val context: Context) : PreferencesDataSt
             }
         }
     }
+
+    override fun clearAll() {
+        runBlocking {
```
newmskywalker, line 69: can we pass a context and launch a coroutine instead? Or Maybe make all the functions in this class suspendable in a different PR? I don't think we should use runBlocking in production

newmskywalker (approved):
🚀

shared/src/androidMain/kotlin/io.newm.shared/internal/implementations/PreferencesDataStoreImpl.android.kt
cristhianescobar: We are not able to pass the context into the method since we the interface is shared across android and iOS, but the Android implementation class does have access to the context so maybe I can use that instead. 
I don't think the `clearAll` needs to be suspended it as is a fire and forget type of operation. We use it when we log out the user.

PR #257, opened 2024-06-13
Dynamically update art work background color

cristhianescobar:


android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryPresenter.kt
```diff
@@ -26,13 +28,18 @@ class NFTLibraryPresenter(
...
+    private val syncWalletConnectionsUseCase: SyncWalletConnectionsUseCase,
     private val walletNFTTracksUseCase: WalletNFTTracksUseCase,
     private val scope: CoroutineScope,
 ) : Presenter<NFTLibraryState> {
     @Composable
     override fun present(): NFTLibraryState {
         val musicPlayer: MusicPlayer? = rememberMediaPlayer()
 
+        LaunchedEffect(Unit) {
+            syncWalletConnectionsUseCase.syncWalletConnectionsFromNetworkToDevice()
```
cristhianescobar, line 40: Sync wallet from network on each app launch

android/features/music-player/src/main/java/io/newm/feature/musicplayer/MusicPlayerViewer.kt
```diff
@@ -305,3 +324,23 @@ fun RepeatButton(
     }
 }
 
+suspend fun getDominantColor(context: Context, imageUri: String): Color {
```
newmskywalker, line 327: Can we add an `onSucess` callback to the existing image request to get the bitmap so we don't have to download the image twice?

newmskywalker (commented):
Love it! 
I think we can leverage the existing image download instead of creating a new request

newmskywalker (approved):

PR #258, opened 2024-06-13
Animate background color change and combine image requests

newmskywalker:
https://github.com/projectNEWM/newm-mobile/assets/3849278/b6a58c7a-1f93-4d54-a832-031110733992


cristhianescobar (approved):
The animation is very smooth!

PR #262, opened 2024-06-20
Swipe to skip song

cristhianescobar:
Adding functionality to the player and mini player to skip to the next song or switch back to the previous song based on swipe gestures on the image in the media player and on the card in the mini player.

### DIFF
--- a/android/core/ui-utils/src/main/java/io/newm/core/ui/utils/SwipeableWrapper.kt
+++ b/android/core/ui-utils/src/main/java/io/newm/core/ui/utils/SwipeableWrapper.kt
@@ -14,19 +14,26 @@ import androidx.compose.ui.unit.dp
 import kotlin.math.roundToInt
 
 /**
+ * Enum class representing the direction of a swipe.
+ */
+enum class SwipeDirection {
+    LEFT, RIGHT
+}
+
+/**
  * A composable function that wraps content and provides swipeable functionality.
  *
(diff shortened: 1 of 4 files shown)


android/core/ui-utils/src/main/java/io/newm/core/ui/utils/SwipeableWrapper.kt
```diff
@@ -0,0 +1,58 @@
...
+ * @param swipeThreshold The distance in Dp that must be swiped to trigger a swipe action.
+ * @param onSwipe Lambda function to be executed when a swipe action is detected.
+ *                The direction parameter is 1 for a left swipe and -1 for a right swipe.
+ * @param content Composable content to be displayed within the swipeable wrapper.
+ */
+@Composable
+fun SwipeableWrapper(
+    modifier: Modifier = Modifier,
+    swipeThreshold: Dp = 100.dp,
+    onSwipe: (direction: Int) -> Unit,
```
newmskywalker: Should we create an enum for the directions?

newmskywalker (approved):
nice!

android/core/ui-utils/src/main/java/io/newm/core/ui/utils/SwipeableWrapper.kt
cristhianescobar: Good idea, I have made the change now

PR #266, opened 2024-06-26
Set default track art icon

cristhianescobar:
Set placeholder and error to default track art.


android/features/music-player/src/main/java/io/newm/feature/musicplayer/MusicPlayerViewer.kt
```diff
@@ -103,6 +103,7 @@ internal fun MusicPlayerViewer(
                 modifier = Modifier.align(Alignment.Center),
                 model = ImageRequest.Builder(context)
                     .data(song.artworkUri)
```
cristhianescobar, line 105: Did not add a placeholder here, because it shows the placeholder every time. Not ideal for the experience we have on the media player.

wlara (approved):

PR #267, opened 2024-06-26
Show loading indicator when nft tracks are being fetched from the network

newmskywalker:


android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryPresenter.kt
```diff
@@ -58,22 +63,22 @@ class NFTLibraryPresenter(
             } else {
                 flowOf()
             }
-        }.collectAsState(initial = emptyList())
+        }.collectAsRetainedState(initial = emptyList())
```
cristhianescobar, line 64: Interesting, so it was not working as expected because we were using `collectAsState` as opposed to `collectAsRetainedState`?

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/HasEmptyWalletUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * This is a hacky way to show a progress indicator
+ * until we know for sure that the user has an empty wallet.
+ *
+ * We need to think about how we want to the handle offline-first functionality
+ * that we currently have. The database is the source of truth, but it's empty while
+ * the NFTs are loading, so we can't rely on it to show the empty state or loading state.
+ */
+interface HasEmptyWalletUseCase {
```
cristhianescobar: Should we merge the functionality with `HasWalletConnectionsUseCase`? Since they are both very similar?

cristhianescobar (changes requested):
Let's combine the usecases since they seem to do the same or similar things.

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/HasEmptyWalletUseCase.kt
newmskywalker: Could someone have a wallet connection without NFTs?

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryPresenter.kt
newmskywalker: it's not working because the database is always empty when we're initially loading the NFTs which results in the UI showing the empty state 

The retained state just ensures we don't make the same call every time the presenter is recreated (similar to how an Android view model would handle this)

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/HasEmptyWalletUseCase.kt
cristhianescobar: Yes, there could be an empty wallet with no NFTs

newmskywalker: Combined it with the `WalletNFTTracksUseCase` since they're more aligned in functionality. Let me know if that works!

cristhianescobar (approved):

PR #270, opened 2024-06-28
Lots of bug fixing

cristhianescobar:
Bug fixing

### DIFF
--- a/android/features/login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountScreenPresenter.kt
+++ b/android/features/login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountScreenPresenter.kt
@@ -7,7 +7,6 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import com.google.android.recaptcha.RecaptchaAction
-import com.google.android.recaptcha.RecaptchaClient
 import com.slack.circuit.retained.rememberRetained
 import com.slack.circuit.runtime.presenter.Presenter
 import io.newm.feature.login.screen.TextFieldState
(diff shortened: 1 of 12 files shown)


wlara (approved):

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/SignupUseCaseImpl.kt
```diff
@@ -10,9 +10,9 @@ import kotlin.coroutines.cancellation.CancellationException
 internal class SignupUseCaseImpl(private val repository: LogInRepository) : SignupUseCase {
 
     @Throws(KMMException::class, CancellationException::class)
-    override suspend fun requestEmailConfirmationCode(email: String, humanVerificationCode: String) {
+    override suspend fun requestEmailConfirmationCode(email: String, humanVerificationCode: String, mustExists: Boolean) {
```
cristhianescobar, line 13: @martyu When calling request `requestEmailConfirmationCode` from the forget password user flow we need to set `mustExists = true`. However, when calling  `requestEmailConfirmationCode` from account creation`mustExists = false`

android/features/login/src/main/java/io/newm/feature/login/screen/createaccount/CreateAccountScreenPresenter.kt
```diff
@@ -69,12 +70,14 @@ class CreateAccountScreenPresenter(
...
                                 step = Step.Loading
                                 step = try {
-                                    recaptchaClientProvider.get().execute(RecaptchaAction.SIGNUP).onSuccess { token ->
-                                        signupUseCase.requestEmailConfirmationCode(
-                                            email = userEmail.text,
-                                            humanVerificationCode = token
-                                        )
-                                    }.onFailure {
+                                    recaptchaClientProvider.get()
+                                        .execute(RecaptchaAction.custom("auth_code"))
```
cristhianescobar, line 74: cc: @martyu When calling the `requestEmailConfirmationCode` we need to ensure we pass in a custom action to recaptcha `auth_code`

PR #273, opened 2024-06-30
Update versions and dependency definition

cristhianescobar:
- Moving the dependency definition closer to what new KMP projects define dependencies
- Fixed iOS built, from previous change from request email code


wlara (approved):

gradle/libs.versions.toml
```diff
@@ -0,0 +1,155 @@
+[versions]
```
wlara, line 1: You can also define your plugins here under `[plugins]`

PR #274, opened 2024-06-30
KMP Room DB support

cristhianescobar:
We want to migrate to Room that is now supported on KMP and away from SQL Delight, given it's easier to mantain and update databases long term.

We have a feature flag that toggles between the current DB implementation and Room.

Before we merge this PR we need to be able to build on both clients:

- [x] Builds on Android
- [ ] Builds on IOS

[Here](https://developer.android.com/kotlin/multiplatform/room) is the documentation for Room if you are curious


shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/NewmPolicyIdsRepository.kt
```diff
@@ -39,7 +39,9 @@ internal class NewmPolicyIdsRepository(
                 isApiCalled = true
             } catch (e: Exception) {
                 // Handle the error appropriately
-                throw e
+                cacheService.getPolicyIds().firstOrNull().let {
+                    emit(it ?: emptyList())
```
wlara: nit: you can do `emit(it.orEmpty())`

shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/WalletRepository.kt
```diff
@@ -17,39 +18,41 @@ internal class WalletRepository(
...
-            connections
-        } catch (e: Exception) {
+            cacheService.getWalletConnections().first()
+        }.getOrElse { e ->
             Logger.e(e) { "Error fetching wallet connections from network ${e.cause}" }
-            return emptyList()
+            cacheService.getWalletConnections().first()
+        }.ifEmpty {
+            emptyList()
         }
```
wlara, line 30: nit: instead of ` ...}.ifEmpty { emptyList() }` you can just do `...}.orEmpty()`

wlara (dismissed):

cristhianescobar:
This PR is waiting on the next version of Room, the issue right now is that it cannot properly link the db implementation on iOS and therefore I am not able to build. But there is a bug reported that should be addressed very soon

https://issuetracker.google.com/issues/342905180?pli=1
cc: @wlara @newmskywalker

cristhianescobar:
Will close this PR since the implementation is way off now, we need a revamp when we migrate Room DB

PR #279, opened 2024-07-08
Client Logger

cristhianescobar:
Adding the ability for the client to register a logger, so KMP and The client code can both use the same logger

- For Android I have added sentry to report errors to

### DIFF
--- a/shared/src/commonMain/kotlin/io.newm.shared/di/NetworkClientFactory.kt
+++ b/shared/src/commonMain/kotlin/io.newm.shared/di/NetworkClientFactory.kt
@@ -9,7 +9,6 @@ import io.ktor.client.plugins.auth.providers.BearerTokens
 import io.ktor.client.plugins.auth.providers.bearer
 import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.client.plugins.defaultRequest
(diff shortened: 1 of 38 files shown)


shared/src/commonMain/kotlin/io.newm.shared/di/NetworkClientFactory.kt
```diff
@@ -72,15 +71,17 @@ internal class NetworkClientFactory(
...
     private val refreshTokenPlugin = createClientPlugin("RefreshTokenPlugin") {
         onRequest { request, _ ->
-            if(request.url.encodedPath == "/v1/auth/refresh") {
+            if (request.url.encodedPath == "/v1/auth/refresh") {
                 request.headers.remove(HttpHeaders.Authorization)
-                request.headers.append(HttpHeaders.Authorization, "Bearer ${tokenManager.getRefreshToken()}")
+                request.headers.append(
+                    HttpHeaders.Authorization,
+                    "Bearer ${tokenManager.getRefreshToken()}"
+                )
```
wlara: nit: I think you can just do `request.bearerAuth(tokenManager.getRefreshToken())`

wlara (approved):

PR #287, opened 2024-07-17
MOBILE-231 Remove nickname flow from registration

cristhianescobar:
The mobile app will not set nickname/stage name. We will only allow user to update their name if they have not KYC, and password.


cristhianescobar:
Heads up @martyu, we will need to remove the nickname flow in registration, as it’s a field that should only be updatable from Newm Studio, given it’s what they referred to as the stage name.

We will also need to remove it from edit profile.

wlara (approved):

PR #292, opened 2024-07-21
Revert circuit upgrade

newmskywalker:


cristhianescobar (approved):
Do we need to rework our navigation before updating circuit?

PR #295, opened 2024-07-21
Replace user's nickname with name and last name

cristhianescobar:


android/app-newm/src/main/java/io/newm/screens/profile/ProfileForm.kt
```diff
@@ -45,13 +47,37 @@ fun ProfileForm(
...
+                if (canUserEditName) {
+                    Email(
+                        label = R.string.profile_form_first_name,
+                        emailState = firstName,
+                        keyboardOptions = KeyboardOptions(
+                            imeAction = ImeAction.Next,
+                        )
                     )
-                )
+                    Email(
```
cristhianescobar, line 58: I will make a ticket to create an actual input component for non email fields. For now we will keep this 😓

shared/src/commonMain/kotlin/io.newm.shared/public/models/User.kt
```diff
@@ -62,3 +62,11 @@ data class User(
     val newPassword: String? = null,
     val confirmPassword: String? = null
 )
+
+fun User.fullName(): String? {
+    return if(!firstName.isNullOrBlank() && !lastName.isNullOrBlank()) "$firstName $lastName" else null
```
wlara, line 67: Some users may have set only one of first name or last name. In the server we use this to support all possibilities when building fullName:
```
"${firstName.orEmpty().trim()} ${lastName.orEmpty().trim()}".trim()
```

wlara (approved):

PR #297, opened 2024-07-22
App Analytics Implementation

cristhianescobar:
Added a framework to track analytics, and evens related to user interactions. 

Here are the list of events the analytics team wants to track:
https://docs.google.com/spreadsheets/d/1OJ7H_hyh3CYDkeIgDNZyBz8Fzq__IrUSR2_8JQAR9h8/edit?gid=0#gid=0

Next: 
I will be tracking all the user interactions

### DIFF
--- a/android/app-newm/src/main/java/io/newm/utils/AndroidNewmAppAnalyticsTracker.kt
+++ b/android/app-newm/src/main/java/io/newm/utils/AndroidNewmAppAnalyticsTracker.kt
@@ -1,6 +1,7 @@
 package io.newm.utils
 
 import android.os.Bundle
+import androidx.core.os.bundleOf
 import com.google.firebase.analytics.FirebaseAnalytics
 import com.google.firebase.analytics.ktx.analytics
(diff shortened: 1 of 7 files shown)


android/app-newm/src/main/java/io/newm/utils/AndroidNewmAppAnalyticsTracker.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    override fun trackEvent(eventName: String, properties: Map<String, Any?>?) {
+        val safeEventName = validateEventName(eventName)
+        if (safeEventName == null) {
+            logger.info(TAG, "Invalid event name: $eventName. Event not tracked.")
+            return
+        }
+
+        val bundle = Bundle().apply {
```
wlara: There is an Android Ktx function that already does this. You can simply do here:
```
val bundle = bundleOf(*properties?.toList()?.toTypedArray().orEmpty()) 
```

wlara (approved):

PR #302, opened 2024-08-04
Add Store5 and create a store for nft tracks

newmskywalker:
https://mobilenativefoundation.github.io/Store/store/store/


cristhianescobar (approved):
LGTM

martyu:
What's the difference in responsibilities between NFTStore and NFTRepository? Looks like they do the same thing?

newmskywalker:
> What's the difference in responsibilities between NFTStore and NFTRepository? Looks like they do the same thing?

Store is just an implementation detail of the repository. 

Notice that I didn't update the repository's contract, and its consumers don't know about Store. The repository is still the centralized place where we can change our caching/fetching strategy. Use cases and presenters don't need to know the details of how store works and what caching strategy to use.

For example, we'll eventually want to add a TTL instead of fetching data on every launch, and managing that will be the repository's responsibility.

PR #304, opened 2024-08-14
Force App Update

cristhianescobar:
- Uptake new Config API changes
- Added the hook to ensure we can force upgrade from login activity or home page

### DIFF
--- a/android/app-newm/build.gradle.kts
+++ b/android/app-newm/build.gradle.kts
@@ -24,7 +24,7 @@ android {
         minSdk = libs.versions.android.minSdk.get().toInt()
         targetSdk = libs.versions.android.targetSdk.get().toInt()
         versionCode = 5
-        versionName = "0.2.5"
+        versionName = "0.3.0"
         testInstrumentationRunner = "io.newm.NewmAndroidJUnitRunner"
         testApplicationId = "io.newm.test"
     }
@@ -57,7 +57,6 @@ android {
             namespace = "io.newm"
             applicationId = "io.newm"
             applicationIdSuffix = ".dev"
-            versionNameSuffix = "-dev"
             dimension = "version"
         }
     }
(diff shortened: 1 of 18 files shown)


android/app-newm/build.gradle.kts
```diff
@@ -24,7 +24,7 @@ android {
         minSdk = libs.versions.android.minSdk.get().toInt()
         targetSdk = libs.versions.android.targetSdk.get().toInt()
         versionCode = 5
```
newmskywalker, line 26: should we be bumping the versionCode as well? 
Maybe we can automate it in a future pr

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ForceAppUpdateUseCaseImpl.kt
```diff
@@ -1,27 +1,35 @@
...
 internal class ForceAppUpdateUseCaseImpl(
     private val remoteConfigRepository: RemoteConfigRepository,
-    private val sharedBuildConfig: NewmSharedBuildConfig
 ) : ForceAppUpdateUseCase {
 
-    override suspend fun isAndroidUpdateRequired(): Boolean =
-        shouldForceAppUpdate { it.android.minAppVersion }
+    override suspend fun isAndroidUpdateRequired(
+        currentAppVersion: String,
+        humanVerificationCode: String
```
newmskywalker, line 14: Are we going to need a captcha for all requests in the future? If so, perhaps we should start thinking about adding the headers using a request interceptor

newmskywalker (approved):
lgtm

shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/ForceAppUpdateUseCaseImpl.kt
cristhianescobar: Up until now we had used captcha for prelogin request, this is the first time we use it on post login flows. We can check with Walter what the roadmap looks like for BE.

android/app-newm/build.gradle.kts
cristhianescobar: Yeah we probably need to automate both version code and string, we can do that on another PR

PR #305, opened 2024-08-14
Fix google sign in on Android

newmskywalker:
We're going to change iOS to support idToken instead of accessToken. I've updated the PR to rename the serializable class to use `idToken` to get Android into a working state in the mean time.

> It appears that Google Sign-In for Android stopped working after we implemented Google Sign-In for iOS. 
> Android needs to send an `idToken`, while iOS should send an `accessToken`. 
> 
> Since both platforms shared the same network request logic, the Android app was serializing the token as `accessToken` instead of the correct `idToken`.
> 
> This PR introduces an expected class to handle the platform-specific tokens, ensuring that Android sends the `idToken` and iOS sends `accessToken`.
> 
> Additionally, I had to update the GOOGLE_AUTH_CLIENT_ID environment variable. However, I wasn't able to test the changes on iOS, so we should verify that iOS functionality still works with the new value before merging this PR.

### DIFF
--- a/shared/src/commonMain/kotlin/io.newm.shared/internal/api/models/Users.kt
+++ b/shared/src/commonMain/kotlin/io.newm.shared/internal/api/models/Users.kt
@@ -9,7 +9,7 @@ internal data class LogInUser(
 )
 
 @Serializable
-internal data class GoogleSignInRequest(val accessToken: String)
+internal data class GoogleSignInRequest(val idToken: String)
 
 @Serializable
 internal data class AppleSignInRequest(val idToken: String)
--- a/shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/LogInRepository.kt
+++ b/shared/src/commonMain/kotlin/io.newm.shared/internal/repositories/LogInRepository.kt
@@ -60,7 +60,7 @@ internal class LogInRepository : KoinComponent {
             when (oAuthData) {
                 is OAuthData.Facebook -> service.loginWithFacebook(FacebookSignInRequest(accessToken = oAuthData.accessToken))
                 is OAuthData.Google -> service.loginWithGoogle(
-                    GoogleSignInRequest(accessToken = oAuthData.idToken),
+                    GoogleSignInRequest(idToken = oAuthData.idToken),
                     humanVerificationCode
                 )
 


cristhianescobar:
@martyu  Can you take a look this PR?

shared/src/androidMain/kotlin/io.newm.shared/internal/api/models/Users.android.kt
```diff
@@ -0,0 +1,6 @@
+package io.newm.shared.internal.api.models
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+actual data class GoogleSignInToken(val idToken: String)
```
cristhianescobar: Should we name the file the same as the data class?

shared/src/iosMain/kotlin/io.newm.shared/internal/api/models/Users.ios.kt
```diff
@@ -0,0 +1,8 @@
+package io.newm.shared.internal.api.models
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+actual data class GoogleSignInToken(
```
cristhianescobar: Same here?

cristhianescobar (dismissed):
LGTM, we just need to rename the files for each of the expect classes

shared/src/commonMain/kotlin/io.newm.shared/internal/api/models/Users.kt
```diff
@@ -9,7 +9,7 @@ internal data class LogInUser(
 )
 
 @Serializable
-internal data class GoogleSignInRequest(val accessToken: String)
+expect class GoogleSignInToken
```
cristhianescobar: I see why the class is name Users, should we move this to it's own class?

android/features/login/src/main/java/io/newm/feature/login/screen/welcome/WelcomeScreenPresenter.kt
```diff
@@ -25,9 +25,9 @@ import io.newm.feature.login.screen.authproviders.RecaptchaClientProvider
 import io.newm.feature.login.screen.authproviders.google.GoogleSignInLauncher
 import io.newm.feature.login.screen.createaccount.CreateAccountScreen
 import io.newm.shared.NewmAppLogger
+import io.newm.shared.internal.api.models.GoogleSignInToken
```
martyu: `internal` classes shouldn't be accessible outside KMM.

shared/build.gradle.kts
```diff
@@ -105,6 +105,14 @@ kotlin {
 			languageSettings.optIn("kotlin.experimental.ExperimentalObjCName")
 		}
 	}
+
```
martyu: Why's this needed?  Didn't we have expect/actuals before?  (`postNotification`)

shared/src/commonMain/kotlin/io.newm.shared/internal/api/models/Users.kt
martyu: This should be `internal` since it's in the API layer.

martyu (changes requested):
Not sure I understand why this is needed...isn't this google sign in call going to *our* backend?  Why is our backend expecting different parameters for ios vs. android?

newmskywalker:
> Not sure I understand why this is needed...isn't this google sign in call going to *our* backend?  Why is our backend expecting different parameters for ios vs. android?

I'll let @wlara chime in regarding the differences, but my understanding is that the new Google sign in sdk uses idToken, so we made it work with either. Not sure if we are able to update iOS to use that instead of accessToken

shared/src/commonMain/kotlin/io.newm.shared/internal/api/models/Users.kt
newmskywalker: Extracted it into its own file and moved it to public. I can't make it internal because constructor invocations will not work in the shared module

shared/build.gradle.kts
newmskywalker: `postNotification` is a function. This is needed for classes

martyu:
I'm able to sign in with google on ios using `accessToken`.  When I change it to `idToken` (both the value and the parameter in the request), the backend returns this error: 

```
{"code":401,"description":"Unauthorized","cause":"Verification failed: The Claim 'aud' value doesn't contain the required audience."}
```


Does Android not work with `accessToken`, but it does work with `idToken`, and you don't get that error?

wlara:
> I'm able to sign in with google on ios using `accessToken`. When I change it to `idToken` (both the value and the parameter in the request), the backend returns this error:
> 
> ```
> {"code":401,"description":"Unauthorized","cause":"Verification failed: The Claim 'aud' value doesn't contain the required audience."}
> ```
> 
> Does Android not work with `accessToken`, but it does work with `idToken`, and you don't get that error?

@martyu Can you send me an ID token value via Slack?  I think that I need to do some backend configuration to allow iOS id tokens.

PR #306, opened 2024-08-16
Make iOS compile against latest shared changes

newmskywalker:


shared/src/commonMain/kotlin/io.newm.shared/internal/implementations/WalletNFTTracksUseCaseImpl.kt
```diff
@@ -38,11 +37,8 @@ internal class WalletNFTTracksUseCaseImpl(
         }
     }
 
-    @Throws(KMMException::class, CancellationException::class)
     override fun getNFTTrack(id: String): NFTTrack? {
-        return mapErrors {
```
martyu, line 43: Whyd this get removed?

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/WalletNFTTracksUseCase.kt
```diff
@@ -57,7 +57,6 @@ interface WalletNFTTracksUseCase {
      * @param id The unique identifier of the NFT track to be retrieved.
      * @return NFTTrack? - The NFT track with the specified ID, or `null` if no such track is found.
      */
-    @Throws(KMMException::class, CancellationException::class)
```
martyu, line 60: Same here

martyu (changes requested):

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/WalletNFTTracksUseCase.kt
newmskywalker: This fixes the error we were talking about on slack. 

image

It looks like throwing functions cannot return optional values. I looked at this specific function and it doesn't throw. It just returns null if the track is not found in the db. So I figured the easiest way to fix the issue was to make the function non-throwing. Let me know if that doesn't work!

newmskywalker: @martyu any additional thoughts on this PR?

cristhianescobar (dismissed):

shared/src/commonMain/kotlin/io.newm.shared/public/usecases/WalletNFTTracksUseCase.kt
cristhianescobar: @martyu are we good to merge this PR?

PR #307, opened 2024-08-18
Analytics Tracker

cristhianescobar:
Track User Analytics screens and button clicks
- [x] Log all screens 
- [ ] Log all button clicks. <-- Will do that on another PR


android/app-newm/src/debug/java/io/newm/DebugNewmApplication.kt
```diff
@@ -22,12 +22,6 @@ class DebugNewmApplication : NewmApplication() {
             addPlugin(CrashReporterPlugin.getInstance())
             addPlugin(DatabasesFlipperPlugin(this@DebugNewmApplication))
             addPlugin(NavigationFlipperPlugin.getInstance())
-            addPlugin(
-                SharedPreferencesFlipperPlugin(
-                    this@DebugNewmApplication,
-                    "newm_encrypted_shared_prefs"
```
cristhianescobar, line 28: Not longer using encrypted share prefs

android/app-newm/src/main/java/io/newm/Logout.kt
```diff
@@ -21,18 +20,20 @@ class Logout(
...
+            }
         }
-        restartApp.run()
-        logger.info("Logout", "Logout successful")
-    } catch (e: Exception) {
-        logger.error("Logout", "Logout failed", e)
     }
 
     fun register() {
-        GlobalScope.launch {
```
cristhianescobar, line 35: GlobalScope is bad

martyu (commented):
Didn't check the android stuff but kmm looks fine.

AndrewWestberg (dismissed):

android/app-newm/src/main/java/io/newm/Logout.kt
```diff
@@ -21,18 +20,20 @@ class Logout(
...
 ) {
 
-    fun signOutUser() = try {
+    fun signOutUser() {
         scope.launch {
-            loginUseCase.logout()
+            try {
+                loginUseCase.logout()
+                restartApp.run()
+                logger.info("Logout", "Logout successful")
```
newmskywalker, line 27: Should we log this before the app restart? Or does the restart return immediately?

android/app-newm/src/main/java/io/newm/screens/forceupdate/ForceAppUpdateUi.kt
```diff
@@ -23,14 +23,17 @@ import io.newm.core.resources.R
...
 
 @Composable
 fun ForceAppUpdateUi(
     state: ForceAppUpdateState,
+    eventLogger: NewmAppEventLogger,
     modifier: Modifier = Modifier
 ) {
     when (state) {
         is ForceAppUpdateState.Content -> {
+            eventLogger.logPageLoad("Force App Update")
```
newmskywalker: I know that this state isn't likely to recompose, but I think we should still launch this in a side-effect

android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditPresenter.kt
```diff
@@ -106,6 +108,7 @@ class ProfileEditPresenter(
         return if (profile == null) {
             ProfileEditUiState.Loading
         } else {
+            eventLogger.logPageLoad("Profile Edit")
```
newmskywalker: This should also happen in a side effect so it doesn't emit on every composition

newmskywalker (commented):
Looks good overall! We do need to launch the page view events in side-effects so they don't get triggered in every composition.

A future enhancement could be to create an analytics side-effect that can be safely launched in a composable.

We could also explore adding the page load events to circuit so that we don't have to do it on every screen

cristhianescobar:
> Looks good overall! We do need to launch the page view events in side-effects so they don't get triggered in every composition.
> 
> A future enhancement could be to create an analytics side-effect that can be safely launched in a composable.
> 
> We could also explore adding the page load events to circuit so that we don't have to do it on every screen

I think once we move all the screens into one activity and we use screen per actual screen rendered it would definitely make sense to move all analytics into the circuit navigation. We could do that as part of the clean up post release. 

I made sure all the analytics are used a side effect it fires ones per screen rendering.

android/app-newm/src/main/java/io/newm/LoginActivity.kt
```diff
@@ -80,16 +82,16 @@ class LoginActivity : ComponentActivity() {
                     CreateAccountUi(state, modifier)
                 }
 
-                is LoginLandingScreen -> ui<WelcomeScreenUiState> { state, modifier ->
-                    WelcomeScreenUi(modifier, state)
+                is Welcome -> ui<WelcomeScreenUiState> { state, modifier ->
+                    WelcomeScreenUi(modifier, state, eventLogger)
```
newmskywalker, line 86: What do you think about injecting the logger into the presenters instead?

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryScreenUi.kt
```diff
@@ -91,22 +95,47 @@ fun NFTLibraryScreenUi(
...
+                }
+                ErrorScreen(state.message)
             }
 
-            NFTLibraryState.EmptyWallet -> EmptyWalletScreen()
-            is NFTLibraryState.Error -> ErrorScreen(state.message)
             is NFTLibraryState.Content -> {
                 val eventSink = state.eventSink
 
+                LaunchedEffect(Unit) {
```
newmskywalker, line 136: Does this log every time the state changes?

wlara (approved):

android/app-newm/src/main/java/io/newm/screens/library/NFTLibraryScreenUi.kt
cristhianescobar: For screens I have tested, only logs once. But buttons I need to do another pass.

android/app-newm/src/main/java/io/newm/LoginActivity.kt
cristhianescobar: That would be ideal but we just need to move some event clicks into the presenter, since not all are there at the moment. Will do that next PR

PR #308, opened 2024-08-25
MOBILE-244 | Implement Account Manager for Android

Jermainelr:
Screenshot_20240825_131145


wlara (approved):

android/app-newm/src/main/java/io/newm/accounts/NewmAccountAuthenticator.kt
```diff
@@ -0,0 +1,50 @@
+package io.newm.accounts
+
+import android.accounts.AbstractAccountAuthenticator
+import android.accounts.Account
+import android.accounts.AccountAuthenticatorResponse
+import android.content.Context
+import android.os.Bundle
+
+internal class NewmAccountAuthenticator(context: Context) : AbstractAccountAuthenticator(context) {
```
cristhianescobar, line 9: We can go with the basic implementation for now, curious to understand what else can be done here @wlara

shared/src/commonMain/kotlin/io.newm.shared/di/Koin.kt
```diff
@@ -90,7 +90,6 @@ fun commonModule(enableNetworkLogs: Boolean) = module {
     single { createJson() }
     // Internal Configurations
     single<NewmSharedBuildConfig> { NewmSharedBuildConfigImpl() }
-    single { TokenManager(get()) }
```
cristhianescobar, line 93: The token manager needs to remain in the KMP layer, as iOS also uses the same interface. With different implementations, curious why did we need to move the injection out of the share module?

cristhianescobar (approved):
Great Job @Jermainelr ! Moving the implementation to the OS makes sense to enable account creation on Android

android/app-newm/src/main/java/io/newm/accounts/NewmAccountAuthenticator.kt
wlara: @cristhianescobar Most of the features are intended for Accounts that go across multiple Apps (e.g., Google User Accounts). For single-app accounts like ours, what we have here is enough.

PR #311, opened 2024-09-06
Analytics Button Clicks

cristhianescobar:
Click Events per screen


shared/src/commonMain/kotlin/io.newm.shared/public/analytics/events/AppScreens.kt
```diff
@@ -5,28 +5,113 @@ import io.newm.shared.public.analytics.ScreenEvents
 
 object AppScreens {
```
cristhianescobar, line 6: @martyu I added all the analytic events to this class in KMP so you can reuse the same screen names and button clicks. Hope this helps

newmskywalker (approved):

PR #318, opened 2024-09-25
Introduced Automatic Versioning with Manual Control Over Major Versions

cristhianescobar:
This PR brings in automatic versioning to simplify our process and keep things consistent.

Automatic `versionCode`: Now generated based on the current date and hour in the format `yyMMddHH`. Each build gets a unique `versionCode`, so you won’t have to worry about updating it manually anymore.

Automatic `versionName`: The `versionName` follows the format `Major.Year.MMDDHH`. The major version is still set manually, but the rest (year, month, day, and hour) is generated automatically, giving us version names that reflect when the build was created.

Manual major version control: We still have full control over the major version, so we can bump it up as needed for big releases or breaking changes.

### DIFF
--- a/android/app-newm/build.gradle.kts
+++ b/android/app-newm/build.gradle.kts
@@ -131,44 +131,50 @@ sentry {
     telemetry.set(true)
 }
 
+
 /**
- * Generates a dynamic `versionCode` based on the current date and hour.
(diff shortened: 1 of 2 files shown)


wlara (approved):

wlara (dismissed):

newmskywalker (dismissed):
Nice!!

PR #328, opened 2024-10-03
add track duration placeholder

newmskywalker:
We currently show `47:04` as the duration for every track while we load the actual duration. This PR makes it so we show a placeholder instead.

- [x] reuse logic for position/duration

**Before**

[Screen_recording_20241002_210406.webm](https://github.com/user-attachments/assets/f0ffa4e4-4b91-4681-acf0-668e14d097eb)

**After**

[Screen_recording_20241002_205345.webm](https://github.com/user-attachments/assets/df1813a9-e670-4e52-8182-9f678fc96284)


cristhianescobar (approved):
The video still shows the 47:04 will that be gone with this change?

newmskywalker:
> The video still shows the 47:04 will that be gone with this change?

mmm are you looking at the before or after video? It should be `-:--`

cristhianescobar:
> > The video still shows the 47:04 will that be gone with this change?
> 
> mmm are you looking at the before or after video? It should be `-:--`

I was looking at the wrong video! Looks great

PR #329, opened 2024-10-03
Add shadows to make music player elements more readable

newmskywalker:
**Before**

**After**


cristhianescobar (approved):
This makes a really bid difference! Thanks

PR #331, opened 2024-10-05
Fix crash in error logger

newmskywalker:
We're catching exceptions arising from `task.getResult`, but then use `task.result` in the error logger. 
This leads to a crash since we're re-throwing the exception


cristhianescobar (approved):
LGTM, just wondering if you were able to reproduce the crash on sentry.

PR #334, opened 2024-10-11
Remove unnecessary assets

cristhianescobar:


cristhianescobar:
Going to merge this PR, looking to decrease the APK size before we submit another APK to the play store

PR #336, opened 2024-10-20
Add feature flag support using launchdarkly

newmskywalker:


android/app-newm/src/main/java/io/newm/AppLaunchGhostActivity.kt
```diff
@@ -18,18 +28,33 @@ class AppLaunchGhostActivity : ComponentActivity() {
...
+                        userDetailsUseCase.fetchLoggedInUserDetailsFlow().filterNotNull().first()
+                    featureFlagMager.setUserId(user.id)
+                } catch (e: Exception) {
+                    logger.error(
+                        tag,
+                        "Failed to identify feature flag because fetching user details failed.",
+                        e
+                    )
+                } finally {
+                    launchHomeActivity()
```
cristhianescobar, line 43: Nice, so we will always attempt to fetch the FF, but if it fails we will use default values.

shared/src/commonMain/kotlin/io.newm.shared/public/featureflags/FeatureFlags.kt
```diff
@@ -0,0 +1,11 @@
+package io.newm.shared.public.featureflags
+
+interface FeatureFlag {
+    val key: String
+}
+
+object FeatureFlags {
+    object MarketPlace : FeatureFlag {
+        override val key = "streams-marketplace"
```
cristhianescobar, line 9: Should we set a default for this FF? Curious how we will handle the situation where feature flags were not fetched? I assume we use default values, and if so should those be declared here along with the key?

cristhianescobar (approved):
LGTM! 🚀

cristhianescobar:
@newmskywalker can you add me to LD when you get a chance please.

shared/src/commonMain/kotlin/io.newm.shared/public/featureflags/FeatureFlags.kt
newmskywalker: The default value is passed when evaluating the flag. Boolean flags (the only ones we currently support) default to false but this can be changed by the caller to `FeatureFlagManager`

android/app-newm/src/main/java/io/newm/AppLaunchGhostActivity.kt
newmskywalker: Yeah, if we're unable to target this specific user, they will be targeted by the other rules we've defined in launchdarkly

PR #341, opened 2024-12-04
Enables downloading tracks behind a feature flag

newmskywalker:


android/features/music-player/src/main/java/io/newm/feature/musicplayer/service/DownloadManager.kt
```diff
@@ -0,0 +1,28 @@
...
+import android.net.Uri
+import androidx.media3.common.util.UnstableApi
+import androidx.media3.exoplayer.offline.DownloadRequest
+import androidx.media3.exoplayer.offline.DownloadService
+
+interface DownloadManager {
+    fun download(id: String, url: String)
+}
+
+internal class DownloadManagerImpl(
```
cristhianescobar: I pull down the branch and it asked me to make this class public before compiling. Are we able to keep it internal by any chance?

cristhianescobar:
Just tested the app, when in airplane mode I am able to listen to music that was previously downloaded, without the visuals. Do we want to include that work as part of this PR or do you prefer to address that on a follow up PR?

newmskywalker:
> Just tested the app, when in airplane mode I am able to listen to music that was previously downloaded, without the visuals. Do we want to include that work as part of this PR or do you prefer to address that on a follow up PR?

I'll raise a different pr for the progress

escobarcristhian (approved):

cristhianescobar:
@newmskywalker can we merge this PR?

PR #342, opened 2024-12-09
Mobile 352

martyu:
https://projectnewm.atlassian.net/browse/MOBILE-352

### DIFF
--- a/shared/src/iosMain/kotlin/io.newm.shared/internal/implementations/PreferencesDataStoreImpl.ios.kt
+++ b/shared/src/iosMain/kotlin/io.newm.shared/internal/implementations/PreferencesDataStoreImpl.ios.kt
@@ -1,46 +1,40 @@
 package io.newm.shared.internal.implementations
 
+import com.liftric.kvault.KVault
 import io.newm.shared.internal.db.PreferencesDataStore
-import platform.Foundation.NSUserDefaults
 
 class PreferencesDataStoreImpl : PreferencesDataStore {
-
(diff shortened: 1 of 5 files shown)


shared/src/iosMain/kotlin/io.newm.shared/internal/implementations/PreferencesDataStoreImpl.ios.kt
```diff
@@ -1,46 +1,40 @@
 package io.newm.shared.internal.implementations
 
+import com.liftric.kvault.KVault
 import io.newm.shared.internal.db.PreferencesDataStore
-import platform.Foundation.NSUserDefaults
 
 class PreferencesDataStoreImpl : PreferencesDataStore {
-
-    private val userDefaults = NSUserDefaults.standardUserDefaults
+    private val vault: KVault = KVault("newm_ios_preferences")
```
cristhianescobar, line 7: Nice, different implementations per client but same interface/protocol. Love it!

cristhianescobar (approved):

shared/src/iosMain/kotlin/io.newm.shared/internal/implementations/PreferencesDataStoreImpl.ios.kt
martyu: @cristhianescobar KVault doesn't sound familiar...? https://github.com/projectNEWM/newm-mobile/pull/241 😉

PR #347, opened 2025-01-18
MOBILE-342 | Translations to British

LiveMasTomas:
Used Google gemini to see if there were any specific UK phrasing we should use. All the strings are the same as in American english except for the sharing ones. Gemini made some suggestions. Also made `RecordStore` non-translatable per @cristhianescobar 

sharing_1


cristhianescobar (approved):
LGTM

cristhianescobar (approved):
LGTM

PR #351, opened 2025-01-28
Investment Portfolio Screen with Stream Tokens

cristhianescobar:
## Details
Added Investment Portfolio Screen: Accessible from the Account screen.
Feature Gating: The feature is gated by LaunchDarkly. If you need access, add your email to the experiment.
Bottom Bar and Mini Player Restructure: Updated visibility logic to allow each screen to determine whether to display the Bottom Bar and Mini Player.

## Next Steps (To be completed in a future PR):
Wire up the amount of tokens earned from each song.

## Account Screen
Screenshot_20250128-012934

## New Screen
Screenshot_20250128-012533


android/app-newm/src/main/java/io/newm/NewmAppComposable.kt
```diff
@@ -107,7 +107,7 @@ internal fun NewmApp(
             context.startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(url)))
         }, eventLogger)
 
-    val currentRootScreen = backstack.topRecord?.screen
+    val currentRootScreen = backstack.topRecord?.screen as Screen
```
LiveMasTomas: is this type cast secure? should we be using `as?` instead?

android/core/resources/src/main/res/values/strings.xml
```diff
@@ -99,6 +99,7 @@
     <string name="profile_disconnect_wallet_button_label">Disconnect Wallet</string>
     <string name="profile_documents">Documents</string>
     <string name="profile_edit_button_label">Edit Profile</string>
+    <string name="profile_investment_portfolio_button_label">Investment Portfolio</string>
```
LiveMasTomas, line 103: nice 😎

LiveMasTomas (dismissed):
Skeleton LGTM!

PR #357, opened 2025-02-17
Update libraries and plugins

newmskywalker:
I skipped the material library because I ran into issues with icons. Will try that at a later time


cristhianescobar (dismissed):
LGTM!

Last time we upgraded circuit there was a navigation issue we ran into, I don't remember what it was out of the top of my head, but curious if you saw any different behavior this time around?

LiveMasTomas (dismissed):

PR #359, opened 2025-02-18
Add Claimable amount to Stream Token List Screen

cristhianescobar:
# What changed?
Added amount of claimable $NEWM tokens based on wallet.

# What's Next?
- Next PR I will add the amount equivalent to USD and how much each song has earned
- Once we have multiple wallet support I will have to update it to fetch multiple wallet earnings
- UI will probably need to be touched up in the near future
- We will need to think through Claimable process for mobile. 

# Screenshot 
image

### DIFF
--- a/android/app-newm/src/main/java/io/newm/screens/investment/portfolio/InvestmentPortfolioPresenter.kt
+++ b/android/app-newm/src/main/java/io/newm/screens/investment/portfolio/InvestmentPortfolioPresenter.kt
@@ -5,6 +5,7 @@ import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.produceState
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.saveable.rememberSaveable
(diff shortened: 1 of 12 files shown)


android/app-newm/src/main/java/io/newm/screens/investment/portfolio/InvestmentPortfolioPresenter.kt
```diff
@@ -6,48 +6,82 @@ import androidx.compose.runtime.collectAsState
...
+        val coroutineScope = rememberCoroutineScope()
+
+        // State to hold claimable token amount
+        var claimableTokenAmount by remember { mutableStateOf(0L) }
 
         LaunchedEffect(Unit) {
-            syncWalletConnectionsUseCase.syncWalletConnectionsFromNetworkToDevice()
+            val wallets = syncWalletConnectionsUseCase.syncWalletConnectionsFromNetworkToDevice()
+            if (wallets.isNotEmpty()) {
+                coroutineScope.launch {
```
LiveMasTomas: I don't think this is necessary, `LaunchedEffect` provides a coroutine scope in `block()`

android/app-newm/src/main/java/io/newm/screens/investment/portfolio/InvestmentPortfolioUi.kt
```diff
@@ -79,35 +99,34 @@ fun InvestmentPortfolioUi(
                 modifier = Modifier
                     .padding(all = 16.dp)
             ) {
-                Text(text = stringResource(id = R.string.you_have_royalties_to_claim))
-                Text(text = "$45.44")
-                Spacer(modifier = Modifier.height(16.dp))
+                Text(
+                    text = stringResource(id = R.string.you_have_royalties_to_claim),
+                    fontSize = 16.sp,
```
LiveMasTomas: we don't use typography from our theme?

LiveMasTomas (changes requested):
denied at the rim

android/app-newm/src/main/java/io/newm/screens/investment/portfolio/InvestmentPortfolioPresenter.kt
```diff
@@ -6,48 +6,82 @@ import androidx.compose.runtime.collectAsState
...
                 walletNFTTracksUseCase.getAllStreamTokensFlow()
             } else {
-                flowOf()
+                flowOf(emptyList())
             }
         }.collectAsRetainedState(initial = emptyList())
 
         return when {
-            streamTokes.isNotEmpty() -> {
+            streamTokens.isNotEmpty() -> {
```
LiveMasTomas, line 69: Lol nice

LiveMasTomas (dismissed):

android/app-newm/src/main/java/io/newm/screens/investment/portfolio/InvestmentPortfolioPresenter.kt
cristhianescobar: Actually we can get rid of it all by using something like produceState
```
val claimableTokenAmount by produceState(initialValue = 0L) { ... }
```

`produceState` manages coroutines internally – It runs a coroutine that updates value, handling state updates within the composable lifecycle.

android/app-newm/src/main/java/io/newm/screens/investment/portfolio/InvestmentPortfolioUi.kt
cristhianescobar: I switch it up to use the theme

PR #360, opened 2025-03-02
MOBILE-360 | Multiple Wallets Screen set up

LiveMasTomas:
* Translations included
* Moved UI to core

https://github.com/user-attachments/assets/1389c960-b44f-40c3-b800-6c2e85023f05


newmskywalker (approved):
lgtm!

cristhianescobar (approved):
LGTM, I pulled down the branch and I was able to connect 3 wallets, but not from the new screen.  @LiveMasTomas can we make another PR to finish out the removing and adding of new wallets. I think we can use the same flow we have in place now (at least the entry point is reusable)

PR #361, opened 2025-03-04
MOBILE-361 | Profile Picture Update

Jermainelr:
upload

### DIFF
--- a/android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditPresenter.kt
+++ b/android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditPresenter.kt
@@ -16,12 +16,15 @@ import io.newm.core.resources.R
 import io.newm.feature.login.screen.TextFieldState
 import io.newm.feature.login.screen.password.isPasswordValid
 import io.newm.feature.login.screen.password.passwordValidationError
+import io.newm.screens.Screen
 import io.newm.screens.Screen.PrivacyPolicy
 import io.newm.screens.Screen.TermsOfService
 import io.newm.screens.profile.OnBack
 import io.newm.screens.profile.OnBottomSheetVisible
 import io.newm.screens.profile.OnConnectWallet
 import io.newm.screens.profile.OnLogout
+import io.newm.screens.profile.OnRemoveProfilePicture
+import io.newm.screens.profile.OnReplaceProfilePicture
 import io.newm.screens.profile.OnSaveProfile
 import io.newm.screens.profile.OnShowPrivacyPolicy
 import io.newm.screens.profile.OnShowTermsAndConditions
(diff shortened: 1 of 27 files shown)


android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditPresenter.kt
```diff
@@ -31,14 +34,17 @@ import io.newm.shared.public.models.User
...
 import io.newm.shared.public.usecases.UserDetailsUseCase
+import io.newm.utils.toTempFile
 import kotlinx.coroutines.launch
 
 class ProfileEditPresenter(
     private val navigator: Navigator,
     private val hasWalletConnectionsUseCase: HasWalletConnectionsUseCase,
     private val userDetailsUseCase: UserDetailsUseCase,
     private val connectWalletUseCase: ConnectWalletUseCase,
+    private val updateProfilePictureUseCase: UpdateProfilePictureUseCase,
```
cristhianescobar, line 48: Can we move the upload flow to be inside the Edit Profile Screen, so a person who is looking at their account don't accidentally start the edit profile picture unless they are in the profile edit flow (when the user hits save we save their new image if any)? I think we might also need a way to show the user that they can update their profile image, I seen apps use a small plus button which might work. Feel free to do whatever you think it's best here. But overall great work!

cristhianescobar (changes requested):
@Jermainelr looks good! We just need to move the flow to the edit profile screen

android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditPresenter.kt
```diff
@@ -188,6 +194,38 @@ class ProfileEditPresenter(
...
+                                navigator.pop()
+                                navigator.goTo(Screen.EditProfile)
+                            } catch (e: Throwable) {
+                                logger.error("ProfileEditPresenter", "An error occurred", e)
+                                errorMessage = context.getString(R.string.profile_error_message)
+                            }
+                        }
+                    }
+
+                    is OnReplaceProfilePicture -> {
```
LiveMasTomas, line 229: Thoughts on adding some kind of loading ui? We don't want the user thinking that the image upload didn't work, especially if they have a slow network. Could be addressed in a different pr

LiveMasTomas: I watched the gif again, maybe we can start the loading ui sooner?

LiveMasTomas (commented):
Looks good, I agree with @cristhianescobar about moving the flow to only be in the edit profile screen.

android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditPresenter.kt
Jermainelr: @cristhianescobar new flow has been implemented as you reccommended, I added the small plus button icon

android/app-newm/src/main/java/io/newm/screens/profile/edit/ProfileEditPresenter.kt
Jermainelr: @LiveMasTomas I moved the loading to when the save button is clicked so it should be handled now

cristhianescobar:
@Jermainelr can you rebase your PR when you have a chance so we can merge the changes in

LiveMasTomas (approved):
LGTM

PR #365, opened 2025-03-09
MOBILE-360 | Multiple Wallets Screen Logic finish

LiveMasTomas:
Finishing up the logic for the multiple wallets screen - still need to make the empty state

https://github.com/user-attachments/assets/c1ad4982-9351-4ae2-a69c-d28f7f9caccd


android/app-newm/src/main/java/io/newm/screens/profile/view/ProfileButtonUi.kt
```diff
@@ -103,8 +103,12 @@ fun WalletsButton(
     }
 
     val requestPermission = rememberRequestPermissionIntent(
-        onGranted = { /*TODO*/ },
-        onDismiss = { /*TODO*/ })
+        onGranted = {
+            val intent = Intent(context, BarcodeScannerActivity::class.java)
```
LiveMasTomas, line 107: this should fix the bug where if you accepted the permissions nothing would happen

LiveMasTomas:
Let me know how we feel about the bottom sheet as the wallet options menu ui

android/app-newm/src/main/java/io/newm/screens/profile/view/ProfileButtonUi.kt
```diff
@@ -103,8 +103,12 @@ fun WalletsButton(
...
+            launcher.launch(intent)
+        },
+        onDismiss = { /*TODO*/ }
```
cristhianescobar, line 110: Should we do something on dismiss? or can this continue to be a noop?

cristhianescobar:
The bottom sheet UI looks good, we can go ahead and merge this @LiveMasTomas 
I think this screen is missing 2 more things IMO

- On the empty state what if we add a "add wallet" button to trigger the flow, this way we can remove the connect wallet on the account page. 
- On the remove wallet, it would be nice to show the user the details of the wallet they are removing (but this is a nice to have)

I will go ahead and merge this PR for now and the rest can be done on another PR

cristhianescobar (approved):

PR #366, opened 2025-03-10
SXSW POC

cristhianescobar:
-Adding marketplace as a webview to the app
-Moved investment porfolio into the main nav bar


LiveMasTomas:
Can we get screenshots?

PR #368, opened 2025-03-18
Add shared CMP module to main app

newmskywalker:


gradle_include/compose.gradle
newmskywalker, line 1: I tried to add the plugin here but that did not work

cristhianescobar (dismissed):
LGTM 🚀

cristhianescobar (approved):

PR #369, opened 2025-03-18
Remove logger from UI for better separation of concerns

newmskywalker:


android/app-newm/src/main/java/io/newm/HomeActivity.kt
```diff
@@ -120,9 +119,8 @@ class HomeActivity : ComponentActivity() {
 
                 is Screen.EditProfile -> ui<ProfileEditUiState> { state, modifier ->
                     ProfileEditUi(
-                        state = state,
```
newmskywalker, line 123: - [x] We need to add these logs to the presenter before merging

cristhianescobar (commented):
Curious, how do you see the reporting of analytics happening here? Do we want to inject the analytics through a lambda when creating the composable?

newmskywalker:
> Curious, how do you see the reporting of analytics happening here? Do we want to inject the analytics through a lambda when creating the composable? 

The presenters should have access to the analytics logger as well as UI events. We just need to log the analytics in the existing event handlers

cristhianescobar (dismissed):
LGTM

PR #370, opened 2025-03-18
Migrate welcome screen UI to CMP

newmskywalker:
With this PR, the welcome screen on the main android app and the compose app are built with CMP

image


android/core/ui-utils/src/commonMain/composeResources/values/strings.xml
```diff
@@ -0,0 +1,199 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<resources>
+    <string name="account">Account</string>
```
newmskywalker, line 3: - [ ] clean up strings
- [ ] handle translations

sharedfeatures/src/commonMain/kotlin/io/newm/sharedfeatures/welcome/WelcomeUi.kt
```diff
@@ -1,28 +1,127 @@
...
+    ) {
+        Column(
+            modifier = Modifier.fillMaxSize(),
+            verticalArrangement = Arrangement.Center,
+            horizontalAlignment = Alignment.CenterHorizontally
+        ) {
+            Text(
+                text = stringResource(R.string.create_account),
+                fontSize = 16.sp,
+//                fontFamily = inter,
```
newmskywalker: - [x] add fonts to CMP

cristhianescobar (approved):
🔥🔥🔥🔥 LFG @newmskywalker ! 🔥🔥🔥🔥🔥🔥

android/app-newm/src/main/java/io/newm/LoginActivity.kt
```diff
@@ -84,7 +84,7 @@ class LoginActivity : ComponentActivity() {
                 }
 
                 is Welcome -> ui<WelcomeScreenUiState> { state, modifier ->
-                    WelcomeScreenUi(modifier, state, eventLogger)
```
cristhianescobar, line 87: The eventLogger is something an interface we inject in the KMP layer, so in theory if the WelcomeScreenUi is in CMP we should be able to provide it, so far it's looking very good

cristhianescobar (dismissed):

PR #371, opened 2025-04-22
Coil 3 migration

newmskywalker:
* Migrate to coil3 which supports CMP
* Replace Android toasts with a cross-platform snackbar


android/core/ui-utils/src/commonMain/kotlin/io/newm/core/ui/LocalSnackBarHostState.kt
```diff
@@ -0,0 +1,8 @@
+package io.newm.core.ui
+
+import androidx.compose.material.SnackbarHostState
+import androidx.compose.runtime.compositionLocalOf
```
newmskywalker, line 4: oops i missed the rest of the changes

cristhianescobar (approved):

PR #374, opened 2025-05-05
MOBILE-360 | Multiple Wallets Screen Empty Screen

LiveMasTomas:
* Adding capability to add a wallet from the wallets empty screen
* Adding alert dialog to disconnecting all wallets flow, in case user hits that button by accident (totally didn't happen to me)
* Need feedback on final designs and copy

https://github.com/user-attachments/assets/0b9ffddc-2d0b-4a21-94eb-fd207f8488ac


cristhianescobar (approved):
Looks good to me, once we have final designs we can work on the translations @LiveMasTomas

PR #379, opened 2025-07-07
MOBILE-360 | Multiple Wallets design update

LiveMasTomas:
MOBILE-360 | Multiple Wallets design update pt 2

* Starting work for wallet details screen

https://github.com/user-attachments/assets/bc486bce-d94a-4f49-b588-899aa110afc0


android/app-newm/src/main/java/io/newm/screens/Screen.kt
```diff
@@ -8,8 +8,12 @@ import com.slack.circuit.runtime.screen.Screen as CircuitScreen
 @Parcelize
 sealed class Screen(val screenName: String, val showBottomBar: Boolean = false, val showMiniPlayer: Boolean = false) : CircuitScreen {
 
-    data object Wallets : Screen(screenName = AppScreens.AccountScreen.name, showBottomBar = true, showMiniPlayer = true)
-
+    data object Wallets : Screen(screenName = AppScreens.WalletsScreen.name, showBottomBar = true, showMiniPlayer = true)
+    data class WalletDetail(val walletId: String, val walletName: String) : Screen(
+        screenName = AppScreens.WalletDetailScreen.name,
+        showBottomBar = true,
+        showMiniPlayer = true
```
cristhianescobar: Nit: This can both be false since we are 3 levels deep in the navigation hierarchy.

shared/src/commonMain/kotlin/io.newm.shared/public/models/WalletConnection.kt
```diff
@@ -5,6 +5,8 @@ import kotlinx.serialization.Serializable
 @Serializable
 data class WalletConnection(
     val id: String,
+    // val localName: String, do we want to add the custom name here?
+    // val blockchainType: String, how we choose the wallet icon
```
cristhianescobar, line 9: Yes, we have a ticket on the backend side to add both of these items to the wallet connection response 
cc: @wlara

cristhianescobar (approved):
LGTM, once BE updates the wallet connection model we can wire up the rest.

android/app-newm/src/main/java/io/newm/screens/walletdetail/WalletDetailPresenter.kt
```diff
@@ -0,0 +1,83 @@
...
+import io.newm.shared.public.models.WalletConnection
+import io.newm.shared.public.models.mocks.EmptyWallet
+import io.newm.shared.public.usecases.FindWalletConnectionUseCase
+import io.newm.shared.public.usecases.SyncWalletConnectionsUseCase
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.launch
+
+class WalletDetailPresenter(
+    private val navigator: Navigator,
+    private val walletID: String,
```
cristhianescobar, line 23: Should the wallet id should be enough to present the wallet details?

Also what do you think about creating a method in the `WalletNFTTracksUseCase` (or adding a new usecase) to query all 
-getAllStreamTokensByWalletID
-getAllTracksFlowByWalletID 

So we display all the items the wallet might have? All the data we need is already available to us

newmskywalker (approved):

android/app-newm/src/main/java/io/newm/screens/walletdetail/WalletDetailPresenter.kt
LiveMasTomas: I wasn't sure if it would be enough to be honest but i can add those use cases in my next PR!

newmskywalker (approved):

PR #380, opened 2025-08-19
Remove iOS github actions

newmskywalker:
These are not very useful. They rarely work, and we often bypass them.


LiveMasTomas (approved):
will this only affect android? or both platform PRs?

cristhianescobar (approved):

PR #383, opened 2025-09-12
init wallet detail page

LiveMasTomas:
Starting to create wallet detail page

https://github.com/user-attachments/assets/d6fcbdf3-9396-4702-b89c-01c85b29b09d

### DIFF
--- a/android/app-newm/src/main/java/io/newm/screens/profile/view/ProfilePresenter.kt
+++ b/android/app-newm/src/main/java/io/newm/screens/profile/view/ProfilePresenter.kt
@@ -74,9 +74,8 @@ class ProfilePresenter(
         val showRecordStore by featureFlagService.observeFlag(FeatureFlags.ShowRecordStore)
             .collectAsState(initial = FeatureFlags.ShowRecordStore.defaultValue)
 
-//        val showMultiWallets by featureFlagService.observeFlag(FeatureFlags.ShowMultiWallets)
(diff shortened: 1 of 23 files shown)


android/app-newm/src/main/java/io/newm/screens/profile/view/ProfilePresenter.kt
```diff
@@ -75,8 +74,9 @@ class ProfilePresenter(
         val showRecordStore by featureFlagService.observeFlag(FeatureFlags.ShowRecordStore)
             .collectAsState(initial = FeatureFlags.ShowRecordStore.defaultValue)
 
-        val showMultiWallets by featureFlagService.observeFlag(FeatureFlags.ShowMultiWallets)
-            .collectAsState(initial = FeatureFlags.ShowMultiWallets.defaultValue)
+//        val showMultiWallets by featureFlagService.observeFlag(FeatureFlags.ShowMultiWallets)
+//            .collectAsState(initial = FeatureFlags.ShowMultiWallets.defaultValue)
+        val showMultiWallets = true
```
cristhianescobar: Since we have the dev menu, we can still use that to change value in the app

LiveMasTomas: whoopsie

cristhianescobar (approved):

PR #389, opened 2025-12-19
Remove rememberMediaPlayer function

newmskywalker:
# Refactor MediaController instantiation to singleton pattern

## Description
This PR refactors the media controller to prevent it from being created and destroyed with the UI lifecycle. 
This caused unnecessary service reconnections during navigation.

### 1. Architecture: Singleton Pattern
*   **New `MediaSessionConnection` Class:** Created a singleton class responsible for managing the lifecycle of the `MediaController` and its connection to the `MediaService`.
*   **Persistence:** The `MusicPlayerImpl` instance is now held by this singleton, ensuring playback state persists across the entire application lifecycle, regardless of UI navigation.

### 2. Dependency Injection
*   Registered `MediaSessionConnection` as a `single` definition in the Koin module (`Dependencies.kt`), making it available app-wide.

### 3. Refactoring: `rememberMediaPlayer` -> `observeMusicPlayer`
*   **Renamed File:** `RememberMediaPlayer.kt` -> `ObserveMusicPlayer.kt`.
*   **Renamed Function:** Changed `rememberMediaPlayer()` to `observeMusicPlayer()`. This naming accurately reflects that the Composable now only *observes* an existing singleton state rather than creating and remembering a new player instance.
*   **Cleanup:** Removed the unused `eventLogger` parameter from the function signature.

### 4. Updated Call Sites
Refactored the following components to use the new `observeMusicPlayer()` function:
*   `MusicPlayerScreen`
*   `MiniPlayer`
*   `NFTLibraryPresenter`

## Benefits
*   **Performance:** Eliminates expensive `MediaController` reconnection logic on every screen composition.
*   **Separation of Concerns:** Decouples the low-level Media3 connection logic from the UI / Composable layer.


cristhianescobar (approved):
LGTM

PR #392, opened 2026-01-09
Migrate login screen to compose multiplatform

newmskywalker:
With this PR, the login screen is completely rendered by compose multiplatform on Android (including the existing app), web, and desktop


cristhianescobar (approved):
LGTM!

Next we might need to remove the login activity so all the screen live in the same activity which might make easier for us as we support other clients

